PR #1, opened 2017-02-28
fluent.syntax.serializer

stasm:
This PR adds the resource serializer to the `fluent.syntax` module.


fluent/syntax/serializer.py
```diff
@@ -0,0 +1,249 @@
...
+        parts.append(serialize_variant(variant, indent + 1))
+
+    parts.append("\n{}".format(" " * 4 * indent))
+
+    return "".join(parts)
+
+
+def serialize_variant(variant, indent=0):
+    return "\n{}{}[{}]{}".format(
+        " " * (4 * indent - 1),
```
zbraniecki: I'd recommend going for 3 whitespaces here and teh fourth will come from the if_default.

tools/prettify.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/python
```
stasm: Sure, I like that better, actually.

fluent/syntax/serializer.py
stasm: But that doesn't give the same result when `indent` is greater than 1, right?

zbraniecki: oh, sorry, I mean, I'd prefer that non-default variants be `"    " * indent`. and default `*` be `"    " * (indent - 1) + "   *"`.

stasm: I don't understand :( Are you saying that you'd prefer to combine the first `{}{}` in the pattern string into a single one?

zbraniecki: Haha, ok. So, your code does this:
```
key = {
|||||[one] One
||||*[two] Two
}
```

what I'd like to see is:
```
key = {
||||[one] One
|||*[two] Two
}
```

Does it illustrate it better?

stasm: Yes, thanks! Except I'm pretty sure it already does the latter and I have tests to prove it :)  It inserts `(4 * indent - 1)` spaces and then adds `*` or an empty space.

zbraniecki: ugh, f**n operator precedence! Thanks :)

PR #4, opened 2017-03-14
Add setup.py

mathjazz:
Adding `setup.py` (mostly copied from https://github.com/l20n/python-l20n/) to be able to pip install the library.

I need this in Pontoon, so I currently use my fork. Other people will need it, too.


Pike:
Could you add a LICENSE file, too, while you're at it and claim a license on this code?

PR #9, opened 2017-05-09
Rebase the migration tools from python-l20n

stasm:
This is a rebase of https://github.com/l20n/python-l20n/pull/5 and its follow-ups (https://github.com/l20n/python-l20n/pull/10, https://github.com/l20n/python-l20n/pull/11 as well as [changes in @Pike's branch](https://github.com/l20n/python-l20n/pull/12#issuecomment-264155662), adjusted to the changes in the Fluent syntax and AST:

- Placeables are no longer lists.
- `ExternalArgument` takes an `Identifier` instance
- Traits are gone, replaced by attributes.
- `Member` is gone, replaced by `Variant` or `Attribute`.
- `Keyword` is gone, replace by `Symbol`.
- Sections are flat.

Since it already got an r+ back in the day, I'm mostly interested in the review of the first two commits.

My next steps are:

- Fix the parsing of Unicode escapes
- Fix the serialization of leading and trailing whitespace
- Review and start working on the bugs @Pike filed back November: https://bugzilla.mozilla.org/buglist.cgi?bug_id=1321298%2C1321290%2C1321271%2C1321281%2C1321279&list_id=13577100


fluent/migrate/transforms.py
```diff
@@ -0,0 +1,281 @@
...
+
+class LITERAL_FROM(SOURCE):
+    """Create a Pattern with the translation value from the given source."""
+
+    def __call__(self, ctx):
+        source = super(self.__class__, self).__call__(ctx)
+        return LITERAL(source)(ctx)
+
+
+class EXTERNAL(Transform):
```
stasm: I'm considering a number of renames for the transforms, which I'll propose in a follow-up PR. Originally, all transforms except for `SOURCE` return a `Pattern` which makes them suitable for using as values of messages. Also `CONCAT` is implemented such that it can unwrap the expression from the `Pattern` and fold more than one transform into a single `Pattern`.

With a simple change to `CONCAT` it should be possible to do the same conditionally: i.e. unwrap `Patterns` when needed while keeping normal expressions unchanged. I think we'll end up with `EXTERNAL_EXPR` for selectors and replacements and `LITERAL_PATTERN`, `PLURAL_PATTERN` and `CONCAT_PATTERNS` for dealing with patterns.

stasm:
I'll port the first two commits to fluent.js when this lands.

PR #11, opened 2017-05-18
Rename transforms

stasm:
Commonly used Transforms which take source information and return Patterns have been renamed as follows:

    LITERAL_FROM → COPY
    REPLACE_FROM → REPLACE
    PLURALS_FROM → PLURALS

Existing primitive Transforms which take text as input are still useful for defining transform functions for variants of the PLURAL transform. They have been changed as follows:

    LITERAL → COPY_TEXT
    REPLACE → REPLACE_TEXT
    PLURALS → (removed)

In addition, the following three helper Transforms have been added:

    TEXT_ELEMENT, EXTERNAL_ARGUMENT and MESSAGE_REFERENCE

They represent common operations in migrations.  In contrast to the ones above, these three do not evauate to Patterns.  Instead, they evaluate to `TextElement` or `Expression` nodes. The CONCAT transform can now accept these as well.

### DIFF
--- a/fluent/migrate/helpers.py
+++ b/fluent/migrate/helpers.py
@@ -1,0 +1,35 @@
+# coding=utf8
+"""Fluent AST helpers.
+
+The functions defined in this module offer a shorthand for defining common AST
+nodes.
+
+They take a string argument and immediately return a corresponding AST node.
+(As opposed to Transforms which are AST nodes on their own and only return the
+migrated AST nodes when they are evaluated by a MergeContext.) """
+
+from __future__ import unicode_literals
+
+import fluent.syntax.ast as FTL
+
+
+def LITERAL(value):
+    """Create a Pattern with a single TextElement."""
+    elements = [FTL.TextElement(value)]
+    return FTL.Pattern(elements)
+
+
+def EXTERNAL_ARGUMENT(name):
+    """Create an ExternalArgument expression."""
+
+    return FTL.ExternalArgument(
+        id=FTL.Identifier(name)
+    )
+
+
+def MESSAGE_REFERENCE(name):
+    """Create a MessageReference expression."""
+
+    return FTL.MessageReference(
+        id=FTL.Identifier(name)
+    )
--- a/fluent/migrate/transforms.py
+++ b/fluent/migrate/transforms.py
@@ -1,7 +1,72 @@
 # coding=utf8
+"""Migration Transforms.
+
+Transforms are AST nodes which describe how legacy translations should be
+migrated.  They are created inert and only return the migrated AST nodes when
+they are evaluated by a MergeContext.
+
+All Transforms evaluate to Fluent Patterns. This makes them suitable for
+defining migrations of values of message, attributes and variants.  The special
+CONCAT Transform is capable of joining multiple Patterns returned by evaluating
+other Transforms into a single Pattern.  It can also concatenate Fluent
+Expressions, like MessageReferences and ExternalArguments.
+
+The COPY, REPLACE and PLURALS Transforms inherit from Source which is a special
+AST Node defining the location (the file path and the id) of the legacy
+translation.  During the migration, the current MergeContext scans the
+migration spec for Source nodes and extracts the information about all legacy
+translations being migrated. Thus,
+
+    COPY('file.dtd', 'hello')
+
+is equivalent to:
+
+    LITERAL(Source('file.dtd', 'hello'))
+
+where LITERAL is a helper defined in the helpers.py module for creating Fluent
+Patterns from the text passed as the argument.
+
+The LITERAL helper and the special REPLACE_IN_TEXT Transforms are useful for
+working with text rather than (path, key) source definitions.  This is the case
+when the migrated translation requires some hardcoded text, e.g. <a> and </a>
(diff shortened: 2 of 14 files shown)


tests/migrate/test_context_real_examples.py
```diff
@@ -142,67 +142,63 @@ def setUp(self):
...
             FTL.Message(
                 id=FTL.Identifier('delete-all-title'),
-                value=LITERAL_FROM(
+                value=COPY(
                     'aboutDownloads.properties',
                     'downloadAction.deleteAll'
                 )
             ),
             FTL.Message(
                 id=FTL.Identifier('delete-all-message'),
```
stasm: This message is a good example of how the renaming makes writing migrations easier.

stasm:
@Pike, would you like to have a look at this? It's a lot of changes due to the fact that I'm renaming things. I recommend looking in `tests` and `tools` to get a sense of what's changed:

https://github.com/projectfluent/python-fluent/pull/11/files#diff-96e856561fd003737768ecf90df23194L150
https://github.com/projectfluent/python-fluent/pull/11/files#diff-173c79db1d1aa5d947447afe8f351ddbL53
https://github.com/projectfluent/python-fluent/pull/11/files#diff-173c79db1d1aa5d947447afe8f351ddbL53

stasm:
> I'm struggling to tell apart TEXT_ELEMENT, COPY, and COPY_TEXT. In particular because TEXT_ELEMENT is allowed next to COPY, but COPY_TEXT isn't?

They were both allowed and I actually realized that given how CONCAT works, TEXT_ELEMENT was redundant. I removed it.

To answer your other questions I'd like to take a step back and summarize how transforms are currently designed. I'm open to changing this, btw. If this explanation is helpful I'll put it in a README, too.

We need to consider both the input and the output.

### Input

The transforms ending with `_TEXT` take raw text (`type() == str`) as input. In majority of cases this text comes from legacy translations which needs to be defined using the SOURCE operation.

```
COPY_TEXT(SOURCE('file.dtd', 'hello'))
```

For convenience, the versions without `_TEXT` take the `path` and the `key` which will be looked up in the legacy translations stored in the current `MergeContext`.

```
COPY('file.dtd', 'hello')
```

Sometimes, it's still useful to use the `*_TEXT` version of a transform. You may want to put a hardcoded text into the migrated translation (e.g. `<a>` and `</a>` when multiple translations become a single one with a DOM overlay); or you may want to combine PLURALS with REPLACE_TEXT. In this case, each variant of the plural string will be run through a REPLACE_TEXT operation. Those variants are strings, so a `REPLACE(path, key, …)` isn't suitable for them.

### Output

Most transforms evaluate to `FTL.Pattern` so that they can be used as values of messages, attributes and variants. This is by far the most common scenario:

```
msg = FTL.Message(
    FTL.Identifier('hello'),
    value=COPY('file.dtd', 'hello'),
)
```

In case of variants, the PLURALS transform slices the original translation into multiple strings, one per each plural category. REPLACE_TEXT is run on each string and evaluates to an `FTL.Pattern` which is suitable as a value of an `FTL.SelectExpression` variant.

```
msg = FTL.Message(
    FTL.Identifier('delete-all'),
    value=PLURALS(
        'file.dtd',
        'deleteAll',
        EXTERNAL_ARGUMENT('num'),
        lambda variant: REPLACE_TEXT(
            variant,
            { '#1': EXTERNAL_ARGUMENT('num') }
        )
    )
)
```

For convenience, I chose to add two transforms which do not evaluate to `FTL.Pattern` and thus cannot be used as values of messages, attributes and variants. These are MESSAGE_REFERENCE and EXTERNAL_ARGUMENT.

Instead of typing:

```
FTL.ExternalArgument(
    id=FTL.Identifier('foo')
)
```

…you can simply type:

```
EXTERNAL_ARGUMENT('foo')
```

They can be used as selectors in the PLURAL transform, as replacements in REPLACE and REPLACE_TEXT and as elements to CONCAT.

---- 

I'm looking for a good balance between consistency with the Fluent AST and convenience. I hope this explanation was helpful. Now, what should we call all these transforms? :)

fluent/migrate/helpers.py
```diff
@@ -0,0 +1,36 @@
+# coding=utf8
+"""Fluent AST helpers.
+
+The functions defined in this module offer a shorthand for defining common AST
+nodes.
+
+They take a string argument and immediately return a corresponding Fluent AST
+node.  (As opposed to Transforms which are proper AST nodes on their own and
+only return Fluent AST nodes when they are evaluated by a MergeContext.)
+"""
```
Pike: Transforms are proper AST nodes which only return AST nodes when they're evaluated?

proper vs Fluent here ...

fluent/migrate/transforms.py
```diff
@@ -1,7 +1,72 @@
 # coding=utf8
+"""Migration Transforms.
+
+Transforms are proper AST nodes which describe how legacy translations should
+be migrated.  They are created inert and only return Fluent AST nodes when they
+are evaluated by a MergeContext.
```
Pike: ... and here

stasm:
> Naming is hard, but is there a better name for whatever really tells those two apart?

They're actually the same thing! I'll just call them `AST nodes` :)

PR #13, opened 2017-06-05
Add Spans to all Nodes

stasm:
Most AST nodes can now have a Span. Use `FluentParser(with_spans=True)` to enable this behavior.  The `with_annotations` config option to `FluentParser` has been removed. The parser always produces Annotations if necessary now.

### DIFF
--- a/fluent/syntax/ast.py
+++ b/fluent/syntax/ast.py
@@ -4,7 +4,7 @@ import json
 
 
 def to_json(value):
-    if isinstance(value, Node):
+    if isinstance(value, BaseNode):
         return value.to_json()
     if isinstance(value, list):
         return list(map(to_json, value))
@@ -19,25 +19,21 @@ def from_json(value):
             k: from_json(v)
             for k, v in value.items()
             if k != 'type'
-            if k != 'span'
         }
-        node = cls(**args)
-
-        # Spans need to be added via add_span, not __init__.
-        if 'span' in value:
-            span = value['span']
-            # Message and section comments don't have their own spans.
-            if span is not None:
-                node.add_span(span['start'], span['end'])
-
-        return node
+        return cls(**args)
     if isinstance(value, list):
         return list(map(from_json, value))
     else:
         return value
 
 
-class Node(object):
+class BaseNode(object):
+    """Base class for all Fluent AST nodes.
+
+    All productions described in the ASDL subclass BaseNode, including Span and
+    Annotation.  Implements __str__, to_json and traverse.
+    """
+
     def traverse(self, fun):
         """Postorder-traverse this node and apply `fun` to all child nodes.
 
@@ -49,7 +45,7 @@ class Node(object):
 
         def visit(value):
             """Call `fun` on `value` and its descendants."""
-            if isinstance(value, Node):
+            if isinstance(value, BaseNode):
                 return value.traverse(fun)
             if isinstance(value, list):
                 return fun(list(map(visit, value)))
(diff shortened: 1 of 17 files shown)


fluent/syntax/ast.py
```diff
@@ -67,174 +82,175 @@ def visit(value):
...
-            {'type': self.__class__.__name__}
+            {'type': u(self.__class__.__name__)}
         )
         return obj
 
     def __str__(self):
         return json.dumps(self.to_json())
 
 
+class Node(BaseNode):
```
zbraniecki: can you write a very short docstring that BaseNode=(Node, MetaNode) and what are Nodes and MetaNodes?

stasm:
> I'm not happy with the parser changes from a python POV. They look like they're methods, but they're not methods on any class, so self is of undefined type. I think we should make those functions proper methods.

Yeah, sorry for not providing more context here. I want them to be methods but I didn't want to completely destroy the diff by indenting them. My plan is to switch to methods in a single commit after I get the r+.

stasm:
In https://github.com/projectfluent/fluent.js/pull/45 which ports these changes to JS, I already switched to methods defined on the `FluentParser` class.

tests/syntax/fixtures_structure/placeable_at_eol.json
```diff
@@ -3,40 +3,75 @@
...
+              "span": {
+                "type": "Span",
+                "start": 39,
+                "end": 48
+              }
+            },
+            "span": {
+              "type": "Span",
+              "start": 39,
+              "end": 48
```
stasm: Can you explain what is off?

tests/syntax/fixtures_structure/placeable_at_eol.json
```diff
@@ -3,40 +3,75 @@
...
             }
           },
           {
             "type": "TextElement",
-            "value": "\nat the end of line.  The message should\nconsist of three lines of text."
+            "value": "\nat the end of line.  The message should\nconsist of three lines of text.",
+            "span": {
+              "type": "Span",
+              "start": 50,
+              "end": 130
```
stasm: The gap comes form the fact that there is ` }` after the `MessageReference`.

stasm: What do you mean by "aggregate TextElements"?

tests/syntax/fixtures_structure/resource_comment.json
```diff
@@ -3,12 +3,17 @@
   "body": [],
   "comment": {
     "type": "Comment",
+    "annotations": [],
+    "content": "This is a resource wide comment\nIt's multiline",
```
stasm: Yes and yes.

tests/syntax/fixtures_structure/placeable_at_eol.json
Pike: The message reference should cover { foo }, I think, and the identifier foo, so I expect those two to have different spans?

Pike:
> Yeah, sorry for not providing more context here. I want them to be methods but I didn't want to completely destroy the diff by indenting them. My plan is to switch to methods in a single commit after I get the r+.

I wonder if that's a good thing to separate out into an extra commit? Sadly, I think it'd be 
1. make functions member methods
2. do the changes in this patch

which I suspect will be hard to convince git to do nicely for you :-/

stasm:
> (…) which I suspect will be hard to convince git to do nicely for you :-/

It wasn't that hard after all :)

I filed https://github.com/projectfluent/fluent/issues/47 to deal with the placeable spans issue. I don't think it should be blocking us here.

Re-requesting review.

fluent/syntax/parser.py
```diff
@@ -27,7 +51,12 @@ def parse(self, source):
 
             ps.skip_ws_lines()
 
-        return ast.Resource(entries, comment)
+        res = ast.Resource(entries, comment)
+
+        if self.with_spans:
+            res.add_span(0, ps.get_index())
```
Pike: could we just decorate parse() with with_span?

fluent/syntax/ast.py
```diff
@@ -3,8 +3,12 @@
 import json
 
 
+if sys.version_info < (3,):
+    str = unicode
```
Pike: As discussed on irc, can you remove this ...

fluent/syntax/ast.py
```diff
@@ -67,174 +67,177 @@ def visit(value):
 
     def to_json(self):
         obj = {
-            name: to_json(value)
+            str(name): to_json(value)
             for name, value in vars(self).items()
         }
         obj.update(
-            {'type': self.__class__.__name__}
+            {'type': str(self.__class__.__name__)}
```
Pike: and the str() usage here?

Pike:
PS: I suggest to not squash those two commits on landing, I think they're nice to keep separated.

fluent/syntax/parser.py
stasm: `parse` is part of the public API and doesn't take `ps` (the `FTLParserStream` instance) as argument. In fact, it creates it for the purpose of parsing.  `ps` is required to get span positions.

PR #14, opened 2017-06-19
Add BaseNode.equals

stasm:


fluent/syntax/ast.py
```diff
@@ -61,6 +75,48 @@ def visit(value):
...
+
+        if self_keys != other_keys:
+            return False
+
+        keys = sorted(self_keys)
+
+        self_fields = [getattr(self, key) for key in keys]
+        other_fields = [getattr(other, key) for key in keys]
+
+        for key, field1, field2 in izip(keys, self_fields, other_fields):
```
Pike: I'd just iterate over keys, unsorted, and assign field1 and field2 inside the loop

fluent/syntax/ast.py
```diff
@@ -61,6 +75,48 @@ def visit(value):
...
+                    return False
+
+                field_sorting = {
+                    # 'annotations': lambda elem: elem.message,
+                    'attributes': lambda elem: elem.id.name,
+                    'tags': lambda elem: elem.name.name,
+                    'variants': lambda elem: elem.key.name,
+                }
+
+                if key in field_sorting.keys():
```
Pike: Just

    if key in field_sorting:

tests/syntax/test_equals.py
```diff
@@ -0,0 +1,228 @@
...
+                { $num ->
+                    [one] One
+                    [two] Two
+                    [few] Few
+                    [many] Many
+                   *[other] Other
+                }
+        """)
+
+        self.assertTrue(message1.equals(message1))
```
Pike: Does it make sense to test for equality against the identity traversal, too?

If so, add that to the tests in this class?

tests/syntax/test_equals.py
```diff
@@ -0,0 +1,228 @@
...
+                   *[b] B
+                    [a] A
+                }
+        """)
+
+        self.assertTrue(message1.equals(message2))
+        self.assertTrue(message2.equals(message1))
+
+
+class TestSpansEqual(unittest.TestCase):
```
Pike: I guess the name here is older than your inequality test? Rename this to TestWithSpans or so?

tests/syntax/test_equals.py
```diff
@@ -0,0 +1,228 @@
...
+
+        messages = [
+            (parser.parse_entry(a), parser.parse_entry(b))
+            for a, b in strings
+        ]
+
+        for a, b in messages:
+            self.assertTrue(a.equals(b, with_spans=True))
+
+    def test_differ_with_spans(self):
```
Pike: Move this test to the end? Then the four combos of flags to Parser and .equals are all in a row?

PR #16, opened 2017-06-22
Add the ignored_fields kwarg to BaseNode.equals

stasm:


fluent/syntax/ast.py
```diff
@@ -73,7 +73,7 @@ def visit(value):
 
         return fun(node)
 
-    def equals(self, other, with_spans=False):
+    def equals(self, other, ignored_fields=['span']):
```
Pike: AFAIK, the pythonic way of doing this is to make the default None, and then at the top of the body do a

    if ignored_fields is None:
        ignored_fields = ['span']

Not sure why exactly, but that's what I've been told about non-trivial things as default arguments.

tests/syntax/test_equals.py
```diff
@@ -237,4 +237,51 @@ def test_differ_with_spans(self):
...
+            ("foo = Foo", "foo = Foobar"),
+        ]
+
+        messages = [
+            (self.parse_ftl_entry(a), self.parse_ftl_entry(b))
+            for a, b in strings
+        ]
+
+        for a, b in messages:
+            self.assertFalse(a.equals(b, ignored_fields=['value']))
```
Pike: As we're adding this in support of comments, can you add some tests here for that in particular so that we don't regress that?

fluent/syntax/ast.py
stasm: It's only needed when you mutate the list inside of the function. The reason why this is tricky is that the default arguments are parsed once when the function is defined. If you then mutate them in the function's body, you end up mutating them for all future calls, too.

The code here doesn't mutate `ignore_fields` at any point.

stasm: One more thing: I'd like to allow `a.equals(b, ignored_fields=None)` which won't be possible with the `if … is None` check.

Pike: 'k

Issue #15, opened 2017-06-22
Attributes and tags and multiline patterns should allow blank lines before them


stasm:
Currently, none of the following messages parses correctly:

```
foo =


    Mutliline Foo Value


bar1


    .attr = Attr


bar2


    .attr1 = Attr1


    .attr2 = Attr2



bar3 =


    Multiline Bar Value


    .attr = Attr


bar4


    .attr =


        Multiline Attr Value


qux1 = Qux


    #tag


qux2 = Qux


    #tag1


    #tag2
```

stasm:
@zbraniecki I'd love to get your help on this when you're back.

zbraniecki:
ugh... I would love us not to allow for that tbh. It looks very confusing to read and the only cases where I can see anything like that ending up in the code is by mistake.
Let's error on mistake.

@mathjazz , @Pike - thoughts?

Pike:
I could see a use in cases where you have quite a lengthy value in mutliple lines, and then a tag. Like

```
foo = 
    This is just a looong
    piece of text
    over multiple
    lines.
    #longtext
```

would probably be easier on the eye with an empty line.

stasm:
I agree with @Pike. I would also like us to err on the side of least surprise. Throwing a syntax error because of a blank line would be surprising to me.

zbraniecki:
ok. I'm convinced by your arguments.

stasm:
@zbraniecki We need to fix the parsers to actually allow the blank lines :)

zbraniecki:
ok. I'm convinced by your argument.

zbraniecki:
this has been fixed in #20

PR #17, opened 2017-08-17
Add the Placeable node

stasm:
WIP. Implement https://github.com/projectfluent/fluent/pull/52.

This doesn't pass structure tests because the AST changes, obviously. All but one behavior tests pass. I'm trying to find the bug.

### DIFF
--- a/fluent/syntax/parser.py
+++ b/fluent/syntax/parser.py
@@ -38,7 +38,7 @@ class FluentParser(object):
         comment = None
 
         ps = FTLParserStream(source)
-        ps.skip_ws_lines()
+        ps.skip_blank_lines()
 
         entries = []
 
@@ -50,7 +50,7 @@ class FluentParser(object):
             else:
                 entries.append(entry)
 
-            ps.skip_ws_lines()
+            ps.skip_blank_lines()
 
         res = ast.Resource(entries, comment)
 
@@ -61,7 +61,7 @@ class FluentParser(object):
 
     def parse_entry(self, source):
         ps = FTLParserStream(source)
-        ps.skip_ws_lines()
+        ps.skip_blank_lines()
         return self.get_entry_or_junk(ps)
 
     def get_entry_or_junk(self, ps):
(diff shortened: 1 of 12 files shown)


stasm:
I'd like to open this up for your feedback, @zbraniecki, before I fix the tests and port this to fluent.js. Thanks!

fluent/syntax/parser.py
```diff
@@ -349,13 +349,14 @@ def get_number(self, ps):
...
-        ps.skip_ws()
+        ps.skip_inline_ws()
 
         while ps.current():
             ch = ps.current()
 
             if ch == '\n':
-                break
+                if not ps.is_peek_next_line_pattern():
+                    break
```
stasm: @zbraniecki This fixed the parsing of the following FTL:

    key =
        { foo }
        Bar

Before this change, `get_pattern` would `break` after the closing `}`. Does this look right to you?

There's a bit of repetition between this code and `get_text_element`. I wonder if that's OK? An alternative would be to make `get_text_element` return `None` in some cases so that we can `break` in `get_pattern`. However that would be inconsistent with the return values of other `get_` methods, I think?

zbraniecki: I think returning `None` as an option is fine. If you really want to avoid that, it seems that returning a `TextElement` with an empty buffer and then testing against the length of the buffer on the last element also works.
I think I prefer that over double `is_peek_next_line_pattern`, but quite honestly, not by that much. If you find the duplication easier, I'm ok with it.

stasm: Thanks. After some more thought I decided to leave the second `is_peek_next_line_pattern` in, with a comment explaining the end-condition of `get_pattern`'s loop.

Checking for length of the value of the last `TextElement` would mean that we create the element, add spans to it in `with_span` and then discard it. Returning `None` would in turn create a precedent where we need to check the truthiness of nodes returned from `get_` methods, at least in `with_span`.

Another reason to not optimize this code wrt. duplication right now is that a lot of it might change when we fix #15. If it then turns out it's easier to return `None` after all, we can add it.

stasm:
Let's move the review of this to https://github.com/projectfluent/fluent.js/pull/64. Once that lands I'll port it to python.

stasm:
@zbraniecki Would you mind taking another look? It's an almost one-to-one port of the JS implementation, with the exception of the `lineOffset` and `columnOffset` functions which are not available in python-fluent yet.

zbraniecki:
yes, sorry!

PR #19, opened 2017-09-01
Compare locales bug 1382622

Pike:
Add testing for python 2.7 with compare-locales and without. As compare-locales doesn't support py3, we don't do that for 3.6 yet.

The test failures were just in the actual test code, so this fix is low-impact.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
+sudo: false
 language: python
 python:
   - "2.7"
   - "3.5"
   - "nightly"
-script: python -m unittest discover
+install: pip install tox-travis
+script: tox
 notifications:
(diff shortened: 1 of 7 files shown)


Pike:
Sadly we don't use tox in automation, need to dig into travis to actually get test coverage.

Pike:
This works, https://travis-ci.org/projectfluent/python-fluent/builds/270754036?utm_source=github_status&utm_medium=notification.

It does run both 2.7 tests in one worker, but the thing that costs time here is setup anyway, so this doesn't affect wall-clock time.

I'd squash commit 1 and 3 here before landing. I think leaving commit 2 separate makes sense.

PR #22, opened 2017-10-11
Run Structure and Behavior tests in Python 3

### LINKED ISSUE #21: tests don't work in py3.6
I spent a good 2 hours today trying to understand why my patch breaks tests.

The two things that made it really hard were:

1) In py2.7 when json doesn't match expected the diff contains all `u'foo' != 'foo'` which is terribly misleading.
2) py3.6 did not report any errors.

The combination of those two led me onto the path of trying to discover if I changed somehow encoding of the tests, because only py2.7 seemed to be affected.

If we can't fix (1), can we at least fix (2) please?

STR:

1) Change any of the fixture_structure json files, for example span start or end.
2) run tox -e py36

Current result:
No error reported

Expected result:
Error reported

stasm:
Fix #21.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,5 +20,6 @@ setup(name='fluent',
       packages=['fluent', 'fluent.syntax', 'fluent.migrate'],
       package_data={
           'fluent.migrate': ['cldr_data/*']
-      }
-      )
+      },
+      tests_require=['six']
+)
--- a/tox.ini
+++ b/tox.ini
@@ -1,11 +1,12 @@
 [tox]
-envlist = py27, py27-cl, py36
+envlist = py27, py27-cl, py35, py36
 skipsdist=True
 
 [testenv]
 setenv =
     PYTHONPATH = {toxinidir}
 deps =
+    six
     cl: compare-locales
 
 commands=python -m unittest discover
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: python
 python:
   - "2.7"
   - "3.5"
+  - "3.6"
   - "nightly"
 install: pip install tox-travis
 script: tox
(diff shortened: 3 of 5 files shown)


setup.py
```diff
@@ -20,5 +20,6 @@
       packages=['fluent', 'fluent.syntax', 'fluent.migrate'],
       package_data={
           'fluent.migrate': ['cldr_data/*']
-      }
-      )
+      },
+      install_requires=['six']
```
stasm: Do we actually need this if it's only used for testing?

tox.ini
```diff
@@ -1,11 +1,12 @@
 [tox]
-envlist = py27, py27-cl, py36
+envlist = py27, py27-cl, py35
```
stasm: I changed this to Python 3.5 because it's the default in Ubuntu 16.04 LTS onwards. The upcoming 17.10 will change to 3.6 but I think we should try to be compatible with the lowest 3.x version possible.

setup.py
Pike: You can use tests_require, I think.

tox.ini
Pike: There doesn't seem to be any good way to install 3.5 on macs.
Also, I don't think we should drop 3.6 completely.

Practically, requiring both will make one platform fail on macs, and the other on linux, and both test on travis. Good enough?

stasm: Yes :)

stasm:
Looking at https://travis-ci.org/projectfluent/python-fluent/builds/286440149, Travis ran 2.7, 2.7 with compare-locales, 3.5 and 3.7 (as `nightly`).

Pike:
> Looking at https://travis-ci.org/projectfluent/python-fluent/builds/286440149, Travis ran 2.7, 2.7 with compare-locales, 3.5 and 3.7 (as nightly).

I think we need to add 3.6 to .travis.yml. It's weird that the 3.5 python picks up the 3.6 test env in tox-travis.

stasm:
Yup, that fixed it, thanks. https://travis-ci.org/projectfluent/python-fluent/builds/286449582

PR #25, opened 2017-10-30
Bug 1412808 - Remove the LITERAL helper.

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1412808

Its most common use-case can be fullfilled with a simple FTL.TextElement() because CONCAT is also capable of concatenating TextElements and Placables.

Its other use-case (defining a hard-coded value for a Message) is rare and the alternative is not too bad:

    FTL.Pattern([
        FTL.TextElement("Hardcoded value")
    ])

Its third use-case was in REPLACE as the foreach function. That's already covered by the default value of foreach, however.


tests/migrate/test_concat.py
```diff
@@ -157,21 +153,20 @@ def setUp(self):
...
+                FTL.Placeable(EXTERNAL_ARGUMENT('channelname')),
                 COPY(self.strings, 'channel.description.end'),
             )
         )
 
         self.assertEqual(
             evaluate(self, msg).to_json(),
             ftl_message_to_json('''
-                channel-desc = You are on the { $channelname } channel.
+                channel-desc = You are on the { $channelname } channel.\u0020
```
Pike: This smells like anticipating a particular solution to [bug 1374246](https://bugzilla.mozilla.org/show_bug.cgi?id=1374246). Want to remove the trailing whitespace in line 153 instead?

stasm: Good point.

PR #27, opened 2017-11-08
Bug 1366298 - Skip SelectExpression in PLURALS for one plural category

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1366298

Languages with one plural category can have their translations migrated
directly to a single `Pattern` rather than a `SelectExpression` with one
variant. This will only happen when the number of variants in the
legacy translation is also one.

This also adds tests for languages with more categories than en-US.

### DIFF
--- a/fluent/migrate/transforms.py
+++ b/fluent/migrate/transforms.py
@@ -236,6 +236,13 @@ class PLURALS(Source):
         selector = evaluate(ctx, self.selector)
         variants = value.split(';')
         keys = ctx.plural_categories
+
+        # A special case for languages with one plural category. We don't need
+        # to insert a SelectExpression at all for them.
+        if len(keys) == len(variants) == 1:
+            variant, = variants
+            return evaluate(ctx, self.foreach(variant))
+
         last_index = min(len(variants), len(keys)) - 1
 
         def createVariant(zipped_enum):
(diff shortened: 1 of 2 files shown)


fluent/migrate/transforms.py
```diff
@@ -236,6 +236,13 @@ def __call__(self, ctx):
         selector = evaluate(ctx, self.selector)
         variants = value.split(';')
         keys = ctx.plural_categories
+
+        # A special case for languages with one plural category. We don't need
+        # to insert a SelectExpression at all for them.
+        if len(keys) == len(variants) == 1:
```
stasm: The special case is only triggered when _both_ the number of plural categories in CLDR and the number of variants in the legacy translation are `1`.

If the legacy translation has more variants we will still insert a `SelectExpression`: https://github.com/projectfluent/python-fluent/pull/27/files#diff-6d4522d85d5c0eb1cbe1e024d2407986R69

tests/migrate/test_plural.py
```diff
@@ -142,5 +143,81 @@ def test_plural_replace(self):
...
+        self.assertEqual(
+            evaluate(self, self.message).to_json(),
+            ftl_message_to_json('''
+                delete-all = { $num } indirme silinsin mi?
+            ''')
+        )
+
+
+@unittest.skipUnless(PropertiesParser, 'compare-locales required')
+class TestPluralManyCategoriesWithReplace(MockContext):
```
Pike: Can you rename the test? Right now, this is test_plural.TestPluralManyCategoriesWithReplace.test_plural.

One plural is enough, how about TestManyCategories.test_with_replace()?

That said, this test sounds more like a test for the upcoming patch about strictly matching plural forms between toolkit and cldr?

PR #28, opened 2017-11-17
Bug 1321290 - Migrate HTML entities to Unicode characters

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1321290

### DIFF
--- a/tests/migrate/test_source.py
+++ b/tests/migrate/test_source.py
@@ -103,7 +103,7 @@ class TestDTD(MockContext):
         self.strings = parse(DTDParser, '''
             <!ENTITY foo "Foo">
 
-            <!ENTITY unicodeEscape "FOO\\u0020BAR">
+            <!ENTITY unicodeEscape "Foo\\u0020Bar">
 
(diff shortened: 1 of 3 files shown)


stasm:
Ci fails because the [compare-locales patch in bug 1321290](https://bugzilla.mozilla.org/show_bug.cgi?id=1321290#c9) needs to land first.

tests/migrate/test_source.py
```diff
@@ -49,3 +55,83 @@ def test_replace(self):
...
+        self.assertEqual(source(self), '&lt;&#x21E7;&#x2318;K&gt;')
+
+
+@unittest.skipUnless(DTDParser, 'compare-locales required')
+class TestDTD(MockContext):
+    def setUp(self):
+        self.strings = parse(DTDParser, '''
+            <!ENTITY foo "Foo">
+
+            <!ENTITY unicodeEscape "FOO\\u0020BAR">
```
Pike: I'd like this to be actually tested.

PR #30, opened 2017-11-27
Bug 1420225 - Read legacy files when scanning for Sources in transforms

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1420225

This is a simple change which makes it unnecessary to use `maybe_add_localization`—at all.

### DIFF
--- a/fluent/migrate/context.py
+++ b/fluent/migrate/context.py
@@ -139,8 +139,12 @@ class MergeContext(object):
         else:
             self.localization_resources[path] = collection
 
-    def add_transforms(self, path, reference, transforms):
-        """Define transforms for path using reference as template.
+    def add_transforms(self, target, reference, transforms):
+        """Define transforms for target using reference as template.
+
+        `target` is a path of the destination FTL file relative to the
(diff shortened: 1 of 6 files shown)


fluent/migrate/context.py
```diff
@@ -139,7 +139,7 @@ def maybe_add_localization(self, path):
         else:
             self.localization_resources[path] = collection
 
-    def add_transforms(self, path, reference, transforms):
+    def add_transforms(self, target, reference, transforms):
         """Define transforms for path using reference as template.
```
Pike: Want to adjust the comment to change from path to target, too?

stasm: Sure, I considered it but thought that the current comment was still OK. Happy to change it.

Pike: I think an udpate would be good, as right now, `for path` isn't good English, if path isn't the argument.

stasm: Yeah, agreed. I updated the comment and also added two sentences about what those paths are supposed to be relative to.

PR #31, opened 2017-11-30
Test latest revision from hg in addition to pypi

Pike:
Just trying if this works.


stasm:
@Pike Do you want to merge this?

Pike:
Good question. In cases like the syntax changes we just went through, this would have perma-failing tests, that's not helpful? OTH, in times where we don't, it'd be great to have tests.

In that sense I guess the answer is no, we shouldn't merge this. The better way to fix this would be to have a separate test on github's integration, but I can't figure out how ad-hoc. Asked twitter.

PR #34, opened 2018-01-04
Bug 1428000 - Migrate: only annotate affected files

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1428000

### DIFF
--- a/tools/migrate/blame.py
+++ b/tools/migrate/blame.py
@@ -1,8 +1,10 @@
 import argparse
 import json
+from os.path import join
+
+from compare_locales.parser import getParser, Junk
 import hglib
 from hglib.util import b, cmdbuilder
-from compare_locales.parser import getParser, Junk
 
 
 class Blame(object):
@@ -11,9 +13,10 @@ class Blame(object):
         self.users = []
         self.blame = {}
 
-    def main(self):
+    def main(self, file_paths):
         args = cmdbuilder(
-            b('annotate'), self.client.root(), d=True, u=True, T='json')
+            b('annotate'), template='json', date=True, user=True,
+            cwd=self.client.root(), file=map(b, file_paths))
         blame_json = ''.join(self.client.rawcommand(args))
         file_blames = json.loads(blame_json)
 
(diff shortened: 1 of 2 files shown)


tools/migrate/migrate-l10n.py
```diff
@@ -38,6 +37,12 @@ def main(lang, reference_dir, localization_dir, blame, migrations, dry_run):
         # Keep track of how many changesets we're committing.
         index = 0
 
+        # Annotate legacy localization files used as sources by this migration
+        # to preserve attribution of translations.
+        files = ctx.localization_resources.keys()
+        blame = Blame(localization_dir).main(files)
```
Pike: I think we should pass the hglib client to Blame directly. That way, we only have one hg process open instead of ... n + 1 minus whatever gets garbage-collected?
 
Either in this patch or in a follow-up, I think.

tools/migrate/blame.py
```diff
@@ -1,8 +1,10 @@
 import argparse
 import json
+from os.path import join
```
Pike: PS: I prefer to have os.path.join explicit rather than importing join into the global namespace. Just `import os` is good enough. At least that's what more pythonic people have told me to do ;-)

stasm:
Thanks for the review!

> I'm a tad unsure on the abspath vs path change. Can you add more detail on that in the commit message? I think I know why it's OK in blame (cwd=root vs not), but not in blame -> migration.

In blame data, `path` is the path relative to CWD and `abspath` is path relative to the root of the repo. With CWD==root, they're the same and removed `abspath` to reduce the number of different _paths_ used in the code.

> The other question I have is whether we should first collect all the migration contexts, then gather all files, and run blame once?

Annotating now takes under a second which I think is good enough. The code is cleaner now when annotating happens for each migration context separately. Also, batching blames would only help us in the scenario when we run multiple migrations touch the same files... and it would likely shave off a fraction of a second per locale or so. Let's keep this simple for now and revisit as a potential perf optimization if we hit another bottleneck.

PR #37, opened 2018-01-29
Rename private messages to terms

stasm:
Implement https://github.com/projectfluent/fluent/pull/86


fluent/syntax/parser.py
```diff
@@ -221,14 +221,17 @@ def get_message(self, ps, comment):
         if pattern is None and attrs is None:
             raise ParseError('E0005', id.name)
 
+        if id.name.startswith('-'):
+            return ast.Term(id, pattern, attrs, comment)
```
Pike: I guess we should throw here if pattern is None?

Probably also in the js impl? Sorry, only now saw this.

stasm: Thanks @Pike! I added a second commit here and in https://github.com/projectfluent/fluent.js/pull/136.

PR #39, opened 2018-02-02
Bug 1435183 - Fix how files are passed to hg annotate

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1435183


stasm:
@Pike The first commit here address the bug in question. `add_transforms` got pretty long so I decided to refactor it a bit in the second commit. It's mostly cosmetic if you don't feel like reviewing it.

PR #41, opened 2018-02-02
Bug 1434998 - Allow for REPLACE values to be Patterns or PatternElements

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1434998

### DIFF
--- a/tests/migrate/test_replace.py
+++ b/tests/migrate/test_replace.py
@@ -15,6 +15,8 @@ from fluent.migrate.transforms import evaluate, REPLACE
 
 
 class MockContext(unittest.TestCase):
+    maxDiff = None
+
     def get_source(self, path, key):
         # Ignore path (test.properties) and get translations from self.strings
         # defined in setUp.
--- a/tests/migrate/test_replace.py
+++ b/tests/migrate/test_replace.py
@@ -148,6 +148,72 @@ class TestReplace(MockContext):
             ''')
         )
 
+    def test_replace_with_placeable(self):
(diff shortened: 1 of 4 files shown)


stasm:
@zbraniecki This should help with the migration in [bug 1424682](https://bugzilla.mozilla.org/show_bug.cgi?id=1424682). The first commit makes it possible to use Patterns and PatternElements as replacement values in `REPLACE`. Take a look at the changes in `ast.py` and the tests. The second commit is a refactor to make the code easier to understand and more Pythonic.

tests/migrate/test_replace.py
```diff
@@ -148,6 +150,72 @@ def test_replace_last(self):
...
+    def test_replace_with_placeable(self):
+        msg = FTL.Message(
+            FTL.Identifier(u'hello'),
+            value=REPLACE(
+                'test.properties',
+                'hello',
+                {
+                    '#1': FTL.Placeable(
+                        EXTERNAL_ARGUMENT('user')
+                    )
```
zbraniecki: Would it work also with just `EXTERNAL_ARGUMENT('user')`?

stasm: Yes, all other tests in this file are written that way :)

stasm: It works because of https://github.com/projectfluent/python-fluent/pull/41/files/e731e8dfbd62cdc6762a5906473aff0925af1e7d#diff-f684603bb97b0b183acc54fecb78696fR176

PR #45, opened 2018-02-06
Allow CONCAT(*elements) and CONCAT(elements=[…])

stasm:
`CONCAT` used to override `FTL.BaseNode.traverse` and `FTL.BaseNode.to_json` because of its use of the `*elements` varargs in `__init__`.

Overriding `to_json` isn't required because the base mathod already supports lists.

<del>For traverse, I refactored the code slightly to only really override what's absolutely required to reconstruct an instance of CONCAT in traverse. The new BaseNode.get_init_args method returns a tuple of (args, kwargs) which can be used to programmatically create a new instance of the same class.</del>

<ins>For traverse, we allow CONCAT to be constructed with both the *elements varargs (useful when writing migration specs) and with the elements=[…] keyword arg (used by traverse because elements is an attribute of CONCAT instances).</ins>

### DIFF
--- a/fluent/migrate/transforms.py
+++ b/fluent/migrate/transforms.py
@@ -277,43 +277,16 @@ class PLURALS(Source):
 
 
 class CONCAT(Transform):
-    """Concatenate elements of many patterns."""
+    """Create a new Pattern from Patterns, PatternElements and Expressions."""
 
-    def __init__(self, *patterns):
-        self.patterns = list(patterns)
+    def __init__(self, *elements, **kwargs):
+        # We want to support both passing elements as *elements in the
+        # migration specs and as elements=[]. The latter is used by
+        # FTL.BaseNode.traverse when it recreates the traversed node using its
+        # attributes as kwargs.
+        self.elements = list(kwargs.get('elements', elements))
 
     def __call__(self, ctx):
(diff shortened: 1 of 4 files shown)


fluent/migrate/transforms.py
```diff
@@ -277,43 +277,16 @@ def createVariant(zipped_enum):
 
 
 class CONCAT(Transform):
-    """Concatenate elements of many patterns."""
+    """Create a new Pattern from Patterns, PatternElements and Expressions."""
 
-    def __init__(self, *patterns):
-        self.patterns = list(patterns)
+    def __init__(self, *elements):
+        self.elements = list(elements)
```
Pike: I wonder if you could just do

    def __init__(self, *arg_elements, elements=None):
        if elements is None:
            elements = arg_elements
        self.elements = list(elements)

and then you wouldn't need `get_init_args` at all?

stasm: Ah, good idea! `*varargs` must follow all keyword args but something like the following should work:

```py
def __init__(self, *elements, **kwargs):
    kwelements = kwargs.get('elements', None)
    if kwelements is not None:
        self.elements = list(kwelements )
    else:
        self.elements = list(elements)
```

PR #46, opened 2018-02-06
Only expose the basic transforms from fluent.migrate

### LINKED ISSUE #43: Remove MESSAGE_REFERENCE and EXTERNAL_ARGUMENT helpers
The [`helpers.py`](https://github.com/projectfluent/python-fluent/blob/b40264348efe0dbad0565678533c24c21a2c4b0e/fluent/migrate/helpers.py) file currently defines two pseudo-AST nodes which can be used in migration specs:  `MESSAGE_REFERENCE` and `EXTERNAL_ARGUMENT`. I'd like to discuss the general approach regarding such helpers.

On one hand the allow to write more concise AST in migrations. On the other, they introduce abstractions which need to be learned besides the AST itself in order to write and maintain migrations. This in turn results in situations when it's tempting to randomly try different combinations of the helpers to see if they fix an issue with the migration :)

AST being the source of truth, I think there's value in sticking to it exclusively, even at the cost of increased verbosity.

@flodolo, @Pike, @zbraniecki - What do you think?

I already see @zbraniecki not use the helpers in [bug 1424682](https://bugzilla.mozilla.org/show_bug.cgi?id=1424682) in some cases, like here:

```py
FTL.Message(
    id = FTL.Identifier('category-general'),
    attributes = [
        FTL.Attribute(
            FTL.Identifier('tooltiptext'),
            FTL.Pattern(
                elements = [
                    FTL.Placeable(
                        # This could read:
                        # expression=MESSAGE_REFERENCE('pane-general-title')
                        expression = FTL.MessageReference(
                            id = FTL.Identifier(
                                'pane-general-title'
                            )
                        )
                    )
                ]
            )
        )
    ]
)
```

This is also related to what the API of `REPLACE` and `CONCAT` should be. Should we only accept `Patterns` and `PatternElements` (i.e. `TextElements` and `Placeables`)? This would make the API closer to the AST. Or should we accept `Expressions` as well? Compare:

```py
# REPLACE converts the MessageReference expression returned by the helper to a Placeable.
FTL.Message(
    id = FTL.Identifier('help-button-label'),
    value = REPLACE(
        'browser/chrome/browser/preferences/preferences.dtd',
        'helpButton2.label',
        {
            '&brandShortName;': MESSAGE_REFERENCE('-brand-short-name')
        }
    )
)
```

```py
# REPLACE requires an explicit Placeable wrapping the MessageReference expression.
FTL.Message(
    id = FTL.Identifier('help-button-label'),
    value = REPLACE(
        'browser/chrome/browser/preferences/preferences.dtd',
        'helpButton2.label',
        {
            '&brandShortName;': FTL.Placeable(
                MESSAGE_REFERENCE('-brand-short-name')
            )
        }
    )
)
```

```py
# REPLACE requires an explicit Placeable and an explicit MessageReference.
FTL.Message(
    id = FTL.Identifier('help-button-label'),
    value = REPLACE(
        'browser/chrome/browser/preferences/preferences.dtd',
        'helpButton2.label',
        {
            '&brandShortName;': FTL.Placeable(
                FTL.MessageReference(
                    FTL.Identifier('-brand-short-name')
                )
            )
        }
    )
)
```

stasm:
When writing Fluent migration specs, only the transforms like `CONCAT`, `COPY`, `PLURALS` and `REPLACE` are really required. All other exports are either helpers or specific to tools/migrate/migrate-l10n.py.

Fixes #43.

### DIFF
--- a/fluent/migrate/__init__.py
+++ b/fluent/migrate/__init__.py
@@ -1,13 +1,5 @@
 # coding=utf8
 
-from .context import MergeContext                      # noqa: F401
-from .errors import (                                  # noqa: F401
-    MigrationError, NotSupportedError, UnreadableReferenceError
-)
 from .transforms import (                              # noqa: F401
-    Source, COPY, REPLACE_IN_TEXT, REPLACE, PLURALS, CONCAT
+    CONCAT, COPY, PLURALS, REPLACE, REPLACE_IN_TEXT
 )
-from .helpers import (                                 # noqa: F401
-    EXTERNAL_ARGUMENT, MESSAGE_REFERENCE
-)
-from .changesets import convert_blame_to_changesets    # noqa: F401
--- a/tools/migrate/examples/about_dialog.py
+++ b/tools/migrate/examples/about_dialog.py
@@ -1,9 +1,8 @@
 # coding=utf8
 
 import fluent.syntax.ast as FTL
-from fluent.migrate import (
-    CONCAT, EXTERNAL_ARGUMENT, MESSAGE_REFERENCE, COPY, REPLACE
-)
+from fluent.migrate import CONCAT, COPY, REPLACE
+from fluent.migrate.helpers import EXTERNAL_ARGUMENT, MESSAGE_REFERENCE
 
 
 def migrate(ctx):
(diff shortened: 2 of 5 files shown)


stasm:
@zbraniecki This doesn't change the way you've been using `fluent.migrate` so far. Are you OK with these changes?

PR #47, opened 2018-02-07
Require compare-locales in fluent.migrate

stasm:
Make compare-locales a hard requirement for running and testing fluent.migrate which already largely depends on it anyways. This reduces the complexity of imports as well as the burden of deciding if a TestCase depends on compare-locales or not. It also makes it less probable to accidentally skip migration tests with `python -m unittest discover`.

The following commands may be used to run tests:

  - `python setup.py test` - Only syntax tests in tests/syntax will be run. compare-locales is not required.

  - `python -m unittest discover` - All tests will be run. If compare-locales is missing, tests will error.

    It's possible to run only syntax or only migrate tests by specifying the starting director on the CLI:

        python -m unittest discover -s tests/syntax
        python -m unittest discover -s tests/migrate

  - `tox` - Tox will create different envs one if which is py27-cl. It contains compare-locales and is configured to run both tests/syntax and tests/migrate tests. All other tests only run tests/syntax.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,5 +21,6 @@ setup(name='fluent',
       package_data={
           'fluent.migrate': ['cldr_data/*']
       },
-      tests_require=['six']
+      tests_require=['six'],
+      test_suite='tests.syntax'
 )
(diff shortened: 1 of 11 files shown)


setup.py
```diff
@@ -21,5 +21,6 @@
       package_data={
           'fluent.migrate': ['cldr_data/*']
       },
-      tests_require=['six']
+      tests_require=['six'],
```
stasm: I tried adding `compare-locales` here but `python setup.py test` fails with version conflicts when it tries to install it. `compare-locales` requires `fluent==0.4.4` while `python-fluent`'s master is usually newer than that.

PR #49, opened 2018-02-09
 PLURALS should be aware of Gecko ordering of variants

stasm:
WIP for https://bugzilla.mozilla.org/show_bug.cgi?id=1415844.

The goal of this bug is the change the migration code such that when a string like downloadMessage.deleteAll is migrated to Fluent in Lithuanian, the first part of the legacy string is assigned to the [one] variant, the second part is assigned to the [other] variant, and the third part is assigned to the [few] variant.

Additionally, we should make sure we choose the default variant well. Right now, the last form as defined in .properties is used as the default. Instead, I suggest that we use 'other' and if it doesn't exist, we use 'many'. According to https://hg.mozilla.org/l10n/compare-locales/file/tip/compare_locales/plurals.py all Gecko plural forms have at least one of these categories.

### DIFF
--- a/fluent/migrate/cldr.py
+++ b/fluent/migrate/cldr.py
@@ -2,6 +2,7 @@
 
 import pkgutil
 import json
+from compare_locales.plurals import CATEGORIES_BY_LOCALE
 
 
 def in_canonical_order(item):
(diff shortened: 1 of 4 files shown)


Pike:
Closing in favor of #52 .

PR #51, opened 2018-02-15
bug 1438570, roll up known translation and apply transform when last …

Pike:
…dependency trickles in

If the dependencies of a transform come in different changesets,
we need to wait until they're all there.
Only if we don't find them all throughout all our changesets, we
shouldn't transform at all.
We can't start with intermediate transforms, 'cause that'd require
that we can do partial transforms, which is still bug 1321271.

Instead of just aggregating changeset.changes, keep track of
all our existing legacy strings in an extra set. That way,
we only transform files and messages touched by this very string.
That's a performance optimization. If we just went for it and
aggregated the changes and passed them to serialize_changeset,
we'd transform each file we touched before, found the snapshot
didn't change and ignore it. Keeping both datasets separate just
makes this quicker.

### DIFF
--- a/tools/migrate/migrate-l10n.py
+++ b/tools/migrate/migrate-l10n.py
@@ -47,10 +47,17 @@ def main(lang, reference_dir, localization_dir, migrations, dry_run):
             if not path.endswith('.ftl'))
         blame = Blame(client).attribution(files)
         changesets = convert_blame_to_changesets(blame)
+        known_legacy_translations = set()
 
         for changeset in changesets:
-            # Run the migration for the changeset.
-            snapshot = ctx.serialize_changeset(changeset['changes'])
+            changes_in_changeset = changeset['changes']
+            known_legacy_translations.update(changes_in_changeset)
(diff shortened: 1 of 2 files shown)


tools/migrate/migrate-l10n.py
```diff
@@ -47,10 +47,17 @@ def main(lang, reference_dir, localization_dir, migrations, dry_run):
...
         changesets = convert_blame_to_changesets(blame)
+        known_legacy_translations = set()
 
         for changeset in changesets:
-            # Run the migration for the changeset.
-            snapshot = ctx.serialize_changeset(changeset['changes'])
+            changes_in_changeset = changeset['changes']
+            known_legacy_translations.update(changes_in_changeset)
+            # Run the migration for the changeset, with the set of
+            # this and all prior legacy translation.
```
flodolo: nit: translations

Pike: Fixed.

PR #52, opened 2018-02-16
Plurals compare locales bug 1415844

Pike:
Moving over to compare-locales to map gecko to cldr plurals.

The tricky piece here is where to put the default variant in the case where the legacy translation doesn't have enough variants.

Different stages of this PR implement different strategies:

1. use the last variant in cldr order, and make that default
2. copy the last variant in cldr order to the default cldr order, and make that default.


flodolo:
I haven't looked at the code yet, but I like the idea of using our own data source for migrations, since we're migrating Mozilla strings. That would solve the problem for locales that are currently missing from CLDR, and for which we know the plural rule to use (or, at least, we think we do). Minus the question I just [left for compare-locales](https://bugzilla.mozilla.org/show_bug.cgi?id=1432163#c6). 

The counter argument would be that it won't work later in Fluent, but there's a larger discussion happening there (e.g. being able to add our own data on top of ICU/CLDR).

fluent/migrate/transforms.py
```diff
@@ -253,23 +254,34 @@ def __call__(self, ctx):
             variant, = variants
             return evaluate(ctx, self.foreach(variant))
 
-        last_index = min(len(variants), len(keys)) - 1
+        # The default cldr form should be the first we have in DEFAULTS
```
flodolo: nit: CLDR

I'm confused by this comment though: aren't we taking the last from DEFAULTS?

Pike: I had refactored this, and forgot to adjust the comment. Sorry, pushed a follow-up.

fluent/migrate/transforms.py
```diff
@@ -253,23 +254,37 @@ def __call__(self, ctx):
...
+        # If we don't have a variant for that, we'll append one,
+        # using the, in CLDR order, last existing variant in the legacy
+        # translation. That may or may not be the last variant.
+        default_key = [
+            key for key in reversed(self.DEFAULT_ORDER) if key in keys
+        ][0]
+
+        keys_and_variants = zip(keys, variants)
+        keys_and_variants.sort(key=lambda (k, v): self.DEFAULT_ORDER.index(k))
+        if keys_and_variants[-1][0] != default_key:
```
stasm: I find the double indexes (`[-1][0]`) hard to read. I wonder if destructing the last element of `keys_and_variants` would help?

```py
last_key, last_variant = keys_and_variants[-1]
if last_key != default_key:
     keys_and_variants.append((default_key, last_variant))
```

tests/migrate/test_plural.py
```diff
@@ -204,7 +254,8 @@ def test_too_few_variants(self):
                     { $num ->
                         [one] Usunąć plik?
                         [few] Usunąć { $num } pliki?
-                       *[many] Usunąć { $num } plików?
+                        [many] Usunąć { $num } plików?
+                       *[other] Usunąć { $num } plików?
```
stasm: `compare-locales` lists [`one, few, many`](https://hg.mozilla.org/l10n/compare-locales/file/tip/compare_locales/plurals.py#l29) for Polish. I was caught off-guard seeing `other` in this test. Can you change the test setup to `plural_categories = ('one', 'few', 'many')` and the Polish string to `deleteAll=Usunąć plik?;Usunąć #1 pliki?` to simulate too few variants? The comment in `test_too_few_variants` needs adjusting, too.

tests/migrate/test_plural.py
```diff
@@ -74,6 +75,55 @@ def test_plural_too_many_variants(self):
...
+
+    def test_common_order(self):
+        self.assertEqual(
+            evaluate(self, self.message).to_json(),
+            ftl_message_to_json('''
+                delete-all =
+                    { $num ->
+                        [one] Pašalinti #1 atsiuntimą?
+                        [few] Pašalinti #1 atsiuntimų?
+                       *[other] Pašalinti #1 atsiuntimus?
```
stasm: This isn't the right order. `few` should end with `atsiuntimus` and `other` should end with `atsiuntimų`. I don't think it's a good idea to reuse these translations to test different orders.

I think it's sufficient to make this test class about weird ordering. If you wish, we could add another test class for testing proper order, but admittedly this is already tested by `TestPlural`.

tests/migrate/test_plural.py
```diff
@@ -74,6 +75,55 @@ def test_plural_too_many_variants(self):
         )
 
 
+class TestPluralOrder(MockContext):
+    # Plural categories corresponding to Lithuanian (lt).
+    plural_categories = ('one', 'few', 'other')
```
stasm: Please change this to `one, other, few` as per https://hg.mozilla.org/l10n/compare-locales/file/tip/compare_locales/plurals.py#l22.

Pike:
Like so?

Dropping review request on gandalf now that stas is back in business.

PR #53, opened 2018-02-28
Bug 1441799 - Support number variant keys in BaseNode.equals

stasm:
`BaseNode.equals` sorts order-agnostic fields such as attributes and variants in order to ensure the equality comparison works correctly even if the order of the elements differs. The code used to assume all variant keys were `VariantName` instances. We need to support numbers as keys as well.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1441799.

### DIFF
--- a/fluent/syntax/ast.py
+++ b/fluent/syntax/ast.py
@@ -103,18 +103,12 @@ class BaseNode(object):
                 if len(field1) != len(field2):
                     return False
 
-                # These functions are used to sort lists of items for when
-                # order doesn't matter.  Annotations are also lists but they
-                # can't be keyed on any of their fields reliably.
-                field_sorting = {
-                    'attributes': lambda elem: elem.id.name,
-                    'variants': lambda elem: elem.key.name,
-                }
-
-                if key in field_sorting:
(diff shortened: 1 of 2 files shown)


fluent/syntax/ast.py
```diff
@@ -103,18 +103,12 @@ def equals(self, other, ignored_fields=['span']):
...
-                if key in field_sorting:
-                    sorting = field_sorting[key]
-                    field1 = sorted(field1, key=sorting)
-                    field2 = sorted(field2, key=sorting)
+                # Sort elements of order-agnostic fields to ensure the
+                # comparison is order-agnostic as well. Annotations should be
+                # here too but they don't have sorting keys.
+                if key in ('attributes', 'variants'):
+                    field1 = sorted(field1, key=lambda elem: elem.sorting_key)
+                    field2 = sorted(field2, key=lambda elem: elem.sorting_key)
```
stasm: This isn't very DRY, but at least it's readable. I considered using (`attrgetter`](https://docs.python.org/3.6/library/operator.html#operator.attrgetter) but in the end I thought it would have hurt readability too.

PR #54, opened 2018-03-06
Migrate empty translations

stasm:
Migrate empty translations as {""}. See https://bugzil.la/1441942.

<del>Empty plural variants are also kept as {""} to preserve as much of the original intent as possible. Dropping them could result in different behavior when the default variant is displayed instead.</del>

<del>This doesn't address migrating leading and trailing whitespace in non-whitespace-only legacy translations. See https://bugzil.la/1374246.</del>

<ins>See https://github.com/projectfluent/python-fluent/pull/54#issuecomment-371536006.</ins>

### DIFF
--- a/fluent/migrate/transforms.py
+++ b/fluent/migrate/transforms.py
@@ -133,24 +133,43 @@ class Transform(FTL.BaseNode):
     def preserve_whitespace(elements):
         # Handle empty values
         if len(elements) == 0:
-            return [
-                FTL.Placeable(
-                    FTL.StringExpression('')
-                )
-            ]
+            return [FTL.Placeable(FTL.StringExpression(''))]
 
-        # Handle whitespace-only values
-        if len(elements) == 1:
-            element, = elements
-            if isinstance(element, FTL.TextElement) \
-                    and re.match(r'^\s*$', element.value):
-                return [
-                    FTL.Placeable(
-                        FTL.StringExpression(element.value)
-                    )
-                ]
+        re_leading = re.compile(r'^(?P<whitespace>\s+)(?P<text>.*?)$')
+        re_trailing = re.compile(r'^(?P<text>.*?)(?P<whitespace>\s+)$')
 
-        return elements
+        def extract_whitespace(regex, element):
+            '''Extract leading or trailing whitespace from a TextElement.
+
+            Return a tuple of (Placeable, TextElement) in which the Placeable
+            encodes the extracted whitespace as a StringExpression and the
+            TextElement has the same amount of whitespace removed. The
+            Placeable with the extracted whitespace is always returned first.
+            '''
+            match = re.search(regex, element.value)
+            if match:
+                whitespace = match.group('whitespace')
+                empty_expr = FTL.Placeable(FTL.StringExpression(whitespace))
+                if whitespace == element.value:
+                    return empty_expr, None
+                else:
+                    return empty_expr, FTL.TextElement(match.group('text'))
+            else:
+                return None, element
+
+        if isinstance(elements[0], FTL.TextElement):
+            ws, text = extract_whitespace(re_leading, elements[0])
+            elements[:1] = [ws, text]
+
+        if isinstance(elements[-1], FTL.TextElement):
+            ws, text = extract_whitespace(re_trailing, elements[-1])
+            elements[-1:] = [text, ws]
+
+        return [
+            element
+            for element in elements
+            if element is not None
+        ]
 
     @staticmethod
     def pattern_of(*elements):
--- a/fluent/migrate/transforms.py
+++ b/fluent/migrate/transforms.py
@@ -22,13 +22,13 @@ translations being migrated. For instance,
 is equivalent to:
 
     FTL.Pattern([
-        FTL.TextElement(Source('file.dtd', 'hello'))
+        Source('file.dtd', 'hello')
     ])
 
 Sometimes it's useful to work with text rather than (path, key) source
-definitions.  This is the case when the migrated translation requires some
+definitions. This is the case when the migrated translation requires some
 hardcoded text, e.g. <a> and </a> when multiple translations become a single
-one with a DOM overlay. In such cases it's best to use the AST nodes:
+one with a DOM overlay. In such cases it's best to use FTL.TextElements:
 
     FTL.Message(
         id=FTL.Identifier('update-failed'),
@@ -41,11 +41,10 @@ one with a DOM overlay. In such cases it's best to use the AST nodes:
         )
     )
 
-The REPLACE_IN_TEXT Transform also takes text as input, making in possible to
-pass it as the foreach function of the PLURALS Transform.  In this case, each
-slice of the plural string will be run through a REPLACE_IN_TEXT operation.
-Those slices are strings, so a REPLACE(path, key, …) wouldn't be suitable for
-them.
+The REPLACE_IN_TEXT Transform also takes TextElements as input, making it
+possible to pass it as the foreach function of the PLURALS Transform. In the
+example below, each slice of the plural string is converted into a
+TextElement by PLURALS and then run through the REPLACE_IN_TEXT transform.
 
     FTL.Message(
         FTL.Identifier('delete-all'),
@@ -64,16 +63,10 @@ them.
 """
 
 from __future__ import unicode_literals
-import itertools
+import re
 
 import fluent.syntax.ast as FTL
 from .errors import NotSupportedError
-
-
-def pattern_from_text(value):
-    return FTL.Pattern([
-        FTL.TextElement(value)
-    ])
 
 
 def evaluate(ctx, node):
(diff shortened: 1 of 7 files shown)


stasm:
I refactored all of `test_plural.py` to use generic terms like `One`, `Few` and `Many` rather than actual translations. Some tests were duplicates so I removed them.

I also changed the behavior of `PLURALS` for legacy translations which define only a single variant or no variants (`key=`). Due to changes from 104f885c5c29264b7fcef983d9bb738aac4348fd most of the time they used to result in a `SelectExpression` with two variants, one taken from the legacy translation and the other one copied to provide the default. I changed it to not insert the `SelectionExpression` at all.

stasm:
I considered dropping empty legacy variants in `PLURALS` but in the end I decided against it. It's functionally different to define a variant as `{""}` and to drop it. The latter scenario results in the default variant being displayed which is different from the behavior defined by the legacy translation.

It would also require more code and tests to properly handle cases where all variants are empty and presumably need to be dropped, or when the default variant is empty, etc.

stasm:
While this PR adds support for migrating empty and all-whitespace values, leading and trailing whitespace in non-whitespace-only legacy translations are still not supported. That's [bug 1374246](https://bugzilla.mozilla.org/show_bug.cgi?id=1374246). I have an idea for how to fix it; at the same time I think it's much lower priority than this PR and I didn't want to make it any more complex.

Pike:
> I considered dropping empty legacy variants in PLURALS but in the end I decided against it. It's functionally different to define a variant as {""} and to drop it. The latter scenario results in the default variant being displayed which is different from the behavior defined by the legacy translation.

@flodolo, is this what we want? From what I've seen so far, and from what I've read in conversations in bugs, it seems that localizers read `;;` to be "no valid choice in this context"?

flodolo:
> @flodolo, is this what we want? From what I've seen so far, and from what I've read in conversations in bugs, it seems that localizers read ;; to be "no valid choice in this context"?

Even English does that: we have strings starting with ";" because the message is only used for more than one element (e.g. warning on closing multiple tabs). 

I think it's better to not create an empty variant, and fall back to the default. E.g.

```properties
tabs.closeWarningMultiple = ;You are about to close #1 tabs. Are you sure you want to continue?
```

Should be migrated to

```properties
tabs-close-warning = 
        { $tabCount ->            
           *[other] You are about to close { $tabCount } tabs. Are you sure you want to continue?
        }

```

More than

```properties
tabs-close-warning = 
        { $tabCount ->            
            [one] {""}
           *[other] You are about to close { $tabCount } tabs. Are you sure you want to continue?
        }

```


Also, DevTools counter example: if they use English, you'd be creating a bunch of empty variants. With the other approach, you would be creating the first variant available, and fall back to the last one. Both are less than ideal, but the latter is definitely better.

stasm:
```properties
tabs-close-warning = 
    { $tabCount ->            
        [one] {""}
       *[other] You are about to close { $tabCount } tabs. Are you sure you want to continue?
    }
```

I understand that this is not ideal, but it's still a valid FTL and a good translation. OTOH, dropping the empty variant will require handling many more edge-cases. I suggest to either not do it at all or file a follow-up. I don't think I'll have time to work on it before I leave on PTO late next week.

stasm:
I'm trying to find an acceptable version of this PR that I can land tomorrow. Right now, the migration code breaks badly on empty values so I think there's still benefit to landing this soon, even with some limitations.

fluent/migrate/transforms.py
```diff
@@ -92,42 +85,85 @@ def __call__(self, ctx):
...
-        pruned = []
-        # Group elements in contiguous sequences of the same type.
-        for elem_type, elems in itertools.groupby(elements, key=type):
-            if elem_type is FTL.TextElement:
-                # Join adjacent TextElements.
-                text = FTL.TextElement(''.join(elem.value for elem in elems))
-                # And remove empty ones.
-                if len(text.value) > 0:
-                    pruned.append(text)
+    def join_adjacent_elements(elements):
```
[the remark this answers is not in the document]
stasm: I'm OK with doing this but I think this is opposite to the direction the fix to handle leading/trailing whitespace would want to go. Do you want to deprioritize or maybe even wontfix [bug 1374246](https://bugzilla.mozilla.org/show_bug.cgi?id=1374246)?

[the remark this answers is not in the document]
Pike: Wouldn't leading/trailing just be trailing pass in `pattern_of`, cutting leading and trailing whitespace off of TextElement into a Placeable, or prefixing with an empty placeable?

stasm:
The morning me thinks that maybe it wouldn't be so hard to add after all. I'll try to do it after breakfast. If it takes more than an hour, let's move it into a follow-up bug.

fluent/migrate/transforms.py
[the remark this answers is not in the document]
stasm: Yes, but it also needs more handling in `CONCAT` so that `CONCAT(COPY(), COPY())` doesn't produce things like `Foo{" "}Bar` when the first `COPY` has trailing whitespace (or the second one has leading whitespace). I'd like to tackle it in a separate PR.

I'll implement your original suggestion to join `TextElements` and `StringExpressions` alike. This might result in `TextElement("      Foo")` in some rare cases. The serializer currently serializes this with the whitespace (although I'd like to change the behavior in the future and throw: [bug 1397233](https://bugzilla.mozilla.org/show_bug.cgi?id=1397233)) but the parses ignores it, so in the end, the whitespace is lost.

This is actually the status quo for translations with leading or trailing whitespace, like `key = \u0020Foo`. I'll unskip two tests to make it explicit. Implementing your suggestion now will prepare us to solve the leading/trailing whitespace issue uniformly in all transforms, regardless of whether it was part of the legacy translation or got concatenated from another legacy translation.

stasm:
It took me a bit longer, although admittedly the feature of dropping plurals did take around an hour to implement. I spent the rest of the day chasing edge-cases related to text normalization. The end result of this rabbit hole exploration is that I've fixed [bug 1374246](https://bugzilla.mozilla.org/show_bug.cgi?id=1374246) :)

stasm:
@Pike: my plan is to land this as three commits; I'll squash the two first ones together. I'm keeping them separate for now to make the review easier. The first commit is what you reviewed previously.

fluent/migrate/transforms.py
```diff
@@ -92,42 +85,106 @@ def __call__(self, ctx):
...
+
+            Return a tuple of (Placeable, TextElement) in which the Placeable
+            encodes the extracted whitespace as a StringExpression and the
+            TextElement has the same amount of whitespace removed. The
+            Placeable with the extracted whitespace is always returned first.
+            '''
+            match = re.search(regex, element.value)
+            if match:
+                whitespace = match.group('whitespace')
+                empty_expr = FTL.Placeable(FTL.StringExpression(whitespace))
```
Pike: Can this be `white_expr`, as it's not empty?

And yes, I still tend to prefer `{""}  foo` to `{"  "}foo`, in which case this would be an `empty_expr` ;-)

fluent/migrate/transforms.py
```diff
@@ -92,42 +85,106 @@ def __call__(self, ctx):
...
+                if whitespace == element.value:
+                    return empty_expr, None
+                else:
+                    return empty_expr, FTL.TextElement(match.group('text'))
             else:
-                pruned.extend(elems)
-        return pruned
+                return None, element
```
Pike: I'd prefer to have this method return lists of variable length ...

fluent/migrate/transforms.py
```diff
@@ -92,42 +85,106 @@ def __call__(self, ctx):
...
+
+        if isinstance(elements[0], FTL.TextElement):
+            ws, text = extract_whitespace(re_leading, elements[0])
+            elements[:1] = [ws, text]
```
Pike: ... and just inject that list here ....

fluent/migrate/transforms.py
```diff
@@ -92,42 +85,106 @@ def __call__(self, ctx):
...
+
+        if isinstance(elements[-1], FTL.TextElement):
+            ws, text = extract_whitespace(re_trailing, elements[-1])
+            elements[-1:] = [text, ws]
```
Pike: ... and here ...

fluent/migrate/transforms.py
```diff
@@ -92,42 +85,106 @@ def __call__(self, ctx):
...
+
+        return [
+            element
+            for element in elements
+            if element is not None
+        ]
```
Pike: ... and then just `return elements` here.

fluent/migrate/transforms.py
```diff
@@ -92,42 +85,106 @@ def __call__(self, ctx):
...
+
+    @staticmethod
+    def pattern_of(*elements):
+        elements = Transform.flatten_elements(elements)
+        elements = Transform.normalize_text_content(elements)
+        elements = Transform.preserve_whitespace(elements)
```
Pike: Just sugar. The first `elements` is a generator, and then it switches to a list. I'm not opposed to the generator, but `preserve_whitespace` requires elements to actually be a list, so this feels a bit brittle?

Maybe just a comment?

fluent/migrate/transforms.py
```diff
@@ -221,59 +276,76 @@ def __init__(self, path, key, replacements):
...
+        # variant in the legacy translation. That may or may not be the last
+        # variant.
         default_key = [
             key for key in reversed(self.DEFAULT_ORDER) if key in keys
         ][0]
 
-        keys_and_variants = zip(keys, variants)
-        keys_and_variants.sort(key=lambda (k, v): self.DEFAULT_ORDER.index(k))
-        last_key, last_variant = keys_and_variants[-1]
+        # Match keys to legacy forms in order they are defined in
```
Pike: grammar nit, `forms in the order`, add the `the`.

fluent/migrate/transforms.py
```diff
@@ -221,59 +276,76 @@ def __init__(self, path, key, replacements):
...
+        # A special case for legacy translations which don't define any
+        # plural forms.
+        if len(pairs) == 0:
+            return Transform.pattern_of()
+
+        # A special case for languages with one plural category or one legacy
+        # variant. We don't need to insert a SelectExpression for them.
+        if len(pairs) == 1:
+            _, only_form = pairs[0]
+            return evaluate(ctx, self.foreach(only_form))
```
Pike: Does this need a `pattern_of`?

fluent/migrate/transforms.py
stasm: Always returning a 2-tuple makes it easy to switch the order in this particular case here. Note that `extract_whitespace` doesn't know if it's processing the leading or the trailing whitespace.

fluent/migrate/transforms.py
stasm: Good point. How about I rename `flatten_elements` to `chain_elements` which is closer to `itertools.chain`?

fluent/migrate/transforms.py
stasm: Or simply `placeable`?

fluent/migrate/transforms.py
stasm: I'll also move `flatten_elements`, `normalize_text_content` and `preserve_whitespace` to regular functions defined at the top of the file. I only really care about `Transform.pattern_of` being part of the official API because it may be useful in custom `foreach` functions passed to `PLURALS`.

fluent/migrate/transforms.py
stasm: Technically, it doesn't. My motivation was that with it, all* transforms return via `pattern_of`. It makes the code easier to follow for me and also keeps the text normalization logic in one place.

_* I noticed one code path in `PLURALS` which didn't; I'll fix it._

fluent/migrate/transforms.py
stasm: Or maybe it's better to keep it all namespaced under `Transform` after all? @Pike do you have an opinion?

stasm: (Insert a joke about renaming to `smoosh_elements`.)

stasm: Ah, I think I have a better idea. I'll push in a few.

stasm: https://github.com/projectfluent/python-fluent/pull/54/commits/8f9bb30510fba7ec8017aa65a8366193ee69be5a is a refactor of `pattern_of`. I'm afraid rebasing this into the other three commits will be hard, so I might just land it as an independent commit.

PR #55, opened 2018-03-15
Bug 1444999 - Wrap hg commit in try...except

flodolo:


tools/migrate/migrate-l10n.py
```diff
@@ -84,9 +84,12 @@ def main(lang, reference_dir, localization_dir, migrations, dry_run):
...
             if not dry_run:
-                client.commit(
-                    b(message), user=b(author), addremove=True
-                )
+                try:
+                    client.commit(
+                        b(message), user=b(author), addremove=True
+                    )
+                except hglib.error.CommandError as err:
+                    print('  WARNING: hg commit failed ({})'.format(err))
```
stasm: In case of Mercurial errors, I think it would make sense to actually interrupt the migration. Should we check if there's anything to commit before we even print `Committing changeset` in line 85?

flodolo: So far I've only found a case where this happens, and that's me playing with the local files. 

Unless there are other use cases, I'm not sure we should interrupt it? The print should likely be move after the commit

The output currently looks something like this:

```
Running migration bug_1435912_preferences_general_xul for it
  Writing to /Users/flodolo/mozilla/mercurial/l10n_clones/locales/it/browser/browser/preferences/preferences.ftl
    Committing changeset: Bug 1435912 - Migrate Preferences::General XUL to Fluent, part 1.
  Warning: hg commit failed ((1, 'invalid branchheads cache (served): tip differs\nnothing changed', ''))
```

stasm: OK, fair enough :) Can you change the indentation of the warning as follows?

        Committing changeset: Bug 1435912 - Migrate Preferences::General XUL to Fluent, part 1.
          WARNING: hg commit failed…

PR #57, opened 2018-03-27
Refactor merge context bug 1436014

Pike:
This refactors `in_changeset` to its own method so that it can be tested explicitly.

Not refactoring all tests yet, but adding one example test on how to do that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 .tox
 *.pyc
+.eggs/
+fluent.egg-info/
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -11,6 +11,7 @@ commands = python -m unittest discover -s tests/syntax
 [testenv:py27-cl]
 deps =
     six
+    mock
```
stasm: Is this required by the PR?

Pike: No, fixed.

PR #62, opened 2018-04-10
fixes bug 1451450, parse VariantExpressions with a MessageReference inside

Pike:
This switches the props of VariantExpression from having a
string `id` to a `MessageReference` `of`

### DIFF
--- a/fluent/syntax/ast.py
+++ b/fluent/syntax/ast.py
@@ -245,9 +245,9 @@ class AttributeExpression(Expression):
 
 
 class VariantExpression(Expression):
-    def __init__(self, id, key, **kwargs):
+    def __init__(self, of, key, **kwargs):
         super(VariantExpression, self).__init__(**kwargs)
-        self.id = id
+        self.of = of
         self.key = key
 
 
--- a/fluent/syntax/serializer.py
+++ b/fluent/syntax/serializer.py
@@ -235,7 +235,7 @@ def serialize_attribute_expression(expr):
 
 def serialize_variant_expression(expr):
     return "{}[{}]".format(
-        serialize_identifier(expr.id),
+        serialize_identifier(expr.of.id),
         serialize_variant_key(expr.key),
     )
 
--- a/fluent/syntax/parser.py
+++ b/fluent/syntax/parser.py
@@ -518,7 +518,7 @@ class FluentParser(object):
             ps.next()
             key = self.get_variant_key(ps)
             ps.expect_char(']')
-            return ast.VariantExpression(literal.id, key)
+            return ast.VariantExpression(literal, key)
 
         if (ch == '('):
             ps.next()
--- a/tests/syntax/fixtures_structure/term.json
+++ b/tests/syntax/fixtures_structure/term.json
@@ -189,9 +189,17 @@
             "type": "Placeable",
             "expression": {
               "type": "VariantExpression",
-              "id": {
-                "type": "Identifier",
-                "name": "-brand-name",
+              "of": {
+                "type": "MessageReference",
+                "id": {
+                  "type": "Identifier",
+                  "name": "-brand-name",
+                  "span": {
+                    "type": "Span",
+                    "start": 145,
+                    "end": 156
+                  }
+                },
                 "span": {
                   "type": "Span",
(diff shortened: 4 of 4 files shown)


fluent/syntax/serializer.py
```diff
@@ -235,7 +235,7 @@ def serialize_attribute_expression(expr):
 
 def serialize_variant_expression(expr):
     return "{}[{}]".format(
-        serialize_identifier(expr.id),
```
stasm: Can you change this to `serialize_expression(expr.of)`?

fluent/syntax/ast.py
```diff
@@ -245,9 +245,9 @@ def __init__(self, id, name, **kwargs):
 
 
 class VariantExpression(Expression):
-    def __init__(self, id, key, **kwargs):
+    def __init__(self, of, key, **kwargs):
         super(VariantExpression, self).__init__(**kwargs)
-        self.id = id
+        self.of = of
```
stasm: @Pike, @zbraniecki - do you have opinions on what we should call this? It should fit both `MessageReference` (now) as well as as `ExternalArgument` (later).

I looked up a few examples: `a.b` parses as `Attribute(value=Name(id="a"), attr: "b")` in Python and as `MemberExpression(object=Identifier(name="a"), property=Identifier(name="b"))` in JavaScript.

stasm: @Pike, @zbraniecki - do you have opinions on what we should call this? It should fit both `MessageReference` (now) as well as as `ExternalArgument` (later).

I looked up a few examples: `a.b` parses as `Attribute(value=Name(id="a"), attr: "b")` in Python and as `MemberExpression(object=Identifier(name="a"), property=Identifier(name="b"))` in JavaScript.

zbraniecki: no strong opinion, both Python and JS naming scheme seems better to me than `of`.

stasm:
@Pike - I think we should also do this for `AttributeExpressions`, for the same reasons. It also matters for the compare-locales check.

Pike:
> @Pike - I think we should also do this for AttributeExpressions, for the same reasons. It also matters for the compare-locales check.

Actually, I think that'd make it harder to do attribute checks. There's a profound semantic difference between `{close-me}`, `{close-me.label}` and `{close-me.accesskey}`, so I don't think we should take one for the other. If they'd be `MessageReferences`, the check would need to find `MessageReference`, but not if it's in an `AttributeExpression`

fluent/syntax/serializer.py
Pike: Follow-up for this landed.

stasm:
Sorry, I didn't make it clear. By "the same reasons" I meant the fact that we'll need to support `AttributeExpressions` on both `MessageReferences` and `ExternalArguments`. We'll make that change in the implementation of Dynamic References. This PR is making ground for it and I'd like it to do it symmetrically for both `VariantExpressions` and `AttributeExpressions`, please.

For compare-locales it sounds like we should have a separate check for symmetrical `AttributeExpressions` of `Messages` (not `Terms`).

In fact, perhaps we should consider introducing `TermReferences` in a follow-up to make this easier?

stasm:
OK, I see. I'm a bit worried about this fix then. I think it's a sound design for `VariantExpression` and it should be extended to `AttributeExpression` but I see how the latter would make some static analysis harder. Perhaps we need a sister method to `BaseNode.traverse` then, with a slightly different mechanics, e.g. return `True` if the traversal should continue deeper.

fluent/syntax/ast.py
stasm: I'd suggest `object` as it's generic enough to fit a wide range of expressions. Another idea: how about `reference`, or `ref`?

  - `-term[varname]` would parse as `VariantExpression(ref=MessageReference(id=Identifier(name="-term")), key=VariantName(name="varname"))`.

  - `$arg[varname]` would parse as `VariantExpression(ref=ExternalArgument(id=Identifier(name="-term")), key=VariantName(name="varname"))`.

stasm:
Let's decide what the name of the `of` field should be and let's land this to unblock Firefox.

Issue #65, opened 2018-05-12
Implement MessageContext and format method


spookylukey:
In other words, make this Python implementation able to generate translations, as mentioned in #58 

I'm adding this issue as a placeholder for anyone else working on this, because I have started to tackle the problem in the 'implement_format' branch on this repo: https://github.com/django-ftl/python-fluent/tree/implement_format

I am following the Javascript implementation to a large degree, but making it more pythonic, and I'm especially leaning on the Javascript tests.

So far I have only the most basic cases covered (simple text messages, and external number/text arguments interpolated into them), but with a decent start in terms of a framework for the remainder. I don't know when I'll be able to look at this again.

spookylukey:
I found some time, and it didn't take as long as I thought. But I really shouldn't do much more on this now! 

I think I've covered most of the spec now (interpreter only, a compiler to Python would give some nice performance improvements for the common case, but will have to wait). There are some usage docs in the README, which also includes a probably incomplete list of TODO items.

I'm going to submit a PR, even though it is incomplete, as a place to do more discussion on the actual implementation.

spookylukey:
This is merged now!

PR #67, opened 2018-05-15
Implement MessageContext.format

spookylukey:
Implementation of #65 

This is incomplete, but at a reviewable level - the remaining work to be done shouldn't influence the overall design that much.

I'm unlikely to get back to this soon, but leaving this here so that others are aware that a start has been made, and if anyone wants to give feedback then they can.


spookylukey:
I based the API and implementation on the JS version. Here are some implementation notes:

1) The API of JS MessageContext seemed a bit unwieldy:

       var errs = [];
       const msg = ctx.getMessage('a-message-id');
       const val = ctx.format(msg, { arg: 'an argument' }, errs);

   The message is retrieved with `getMessage`, and then just passed straight back to another method on `MessageContext` - it is essentially an opaque object. Plus it requires manually creating the errors argument, and requires three lines to do the job.

   So in Python I changed the API to:

       val, errs = ctx.format('a-message-id', {'arg': 'an argument'})

   Any expected errors that occur while generating the translation are collected in the errors list, as per the Javascript API. If the message itself does not exist at all, however, you will get a LookupError - this seems more in keeping with what Python users would expect.

2) For messages with attributes, the Javascript version is

       const msg = ctx.getMessage('a-message-id').attrs.subattribute;

   This means you have to know something about the type of object returned by `getMessage` (i.e. the `attrs` attribute), but then again you just pass the result back to 'format' to do anything useful.

   So I changed that to:

       val, errs = ctx.format('a-message-id.subattribute', {'arg': 'an argument'})

3) `getMessage` and friends are returning objects that are essentially an implementation detail - AST objects returned by the parser. There is nothing a normal user of MessageContext should be doing with these, and with a different implementation strategy (e.g. compilation instead of interpreting) this might not make any sense.

   So, I've removed these methods. There is still `has_message()` to check a message exists, and `message_idss()` that returns a list of identifiers only, not the contents.

4) In the JS version, the implementation uses 'toString' quite a bit, and relies on the fact that you can call `"a string".toString(context)` and the extra parameter `context` gets ignored, which is the desired behaviour. The equivalent does not hold for Python, so I had to structure the code a bit differently.

   In addition, I wanted this implementation to be as loosely tied to the classes in the parser AST modules as possible, so I've done everything with 'external' polymorphism rather than modify those AST classes and add methods to them. (I've used 'functools.singledispatch', which is really a fancy but
   nicer 'isinstance')

5) JS only has 'Number', in Python we have 'int' and 'float'. This complicates a few things relative to the JS version.

6) I've added a few dependencies. The biggest is Babel (python-babel, not the JS thing), a collection of i18n utilities which has a very nice implementation of plural rules, and includes CLDR data, which made implementation of matching plural categories really easy, and also provides number formatting.

spookylukey:
It turns out the `MessageContext.format` API I've suggested here (accepting an ID rather than a Message object) is also being suggested for the JS version - https://github.com/projectfluent/fluent.js/issues/208

stasm:
Yes :) It was your PR here which made me file that issue. The current API is an artifact of how we localize attributes in the HTML bindings. We might want to keep it in the JS implementation, or maybe change the way the bindings work. I think it merits a discussion which is why I filed the issue. Thanks!

spookylukey:
The biggest remaining TODO on this PR is supporting `DATETIME`, which means implementing `Intl.DateTimeFormat` in Python. Unfortunately this looks like a major obstacle. The spec for it is highly involved, and all written assuming a Javascript context - http://www.ecma-international.org/ecma-402/1.0/#sec-12 . Trying to work out the core of the algorithm compared to the Javascript details, and which bits I actually would need to implement, is not at all easy.

The best options is likely to start with [Intl.js](https://github.com/andyearnshaw/Intl.js/blob/master/src/12.datetimeformat.js), but it still looks like being more work than everything else in this patch put together.

zbraniecki:
I don't think it should block. `DATETIME` is just one of the formatters, and we should aim to get python's API for CLDR datetime formatting for it. Until then, we can either do a dummy date formatting, or skip it all together.

It's important to get it in eventually, but it shouldn't block us from introducing MessageContext API.

zbraniecki:
@spookylukey what prevents you from using http://babel.pocoo.org/en/latest/dates.html ? It seems like this should work quite well, especially the `format` part which resembles the ECMA402 proposed `style` option proposal: https://github.com/tc39/proposal-ecma402-datetime-style

I would recommend not using the pattern/skeleton options, but I think we could start with just implementing `DATETIME` function without params, or with the `format/style` param only.

Later, we could try to add the option bag that finds the matching skeleton/pattern.

stasm:
I'd say it's totally fine to not support `DATETIME` for now. This is already a big PR and a great addition to `python-fluent`! If you agree to leave it out, please file a new issue about adding the support later on.

spookylukey:
@zbraniecki - after a bit of googling the other day I had come across that proposal and had the same idea. I'll implement DATETIME with only the `style/dateStyle/timeStyle` option(s) for now.

The other options should all be do-able with time, and I think python-babel already exports the data needed to do implement them and get the outputs to match `Intl.DateTimeFormat`. It's just it will take a while to dig to through the spec and work out what is actually needed, especially as I'm not expert enough with Javascript object model to know the significance of the different ways of constructing objects (e.g. the spec talks about using `Object.create` and `Record` etc.)

spookylukey:
Regarding docs - we're probably at the point where we need proper docs, rather than just a README. My instinct is to use Sphinx, which I've used lots before and has great support for Python, plus it only takes a few minutes to set up readthedocs.org and get docs built automatically etc. Any objections to me documenting using Sphinx?

spookylukey:
OK, I've implemented DATETIME roughly as discussed. This is ready for review now. I understand that might take some time :-) , I'm also likely to be offline for a while.

spookylukey:
@stasm Another note about this patch: I haven't focussed on performance that much. However, I have another set of patches which add a performance focussed `MessageContext` implementation which compiles FTL to Python, as opposed to the interpreter approach here. That work is nearing completion. It doesn't supersede this PR however - for various reasons that I'll explain later it will be useful to have both implementations.

stasm:
Hey @spookylukey, thanks for your continued work in this PR! I was busy with the updates to the Syntax Spec; now that 0.6 is out, I hope to be able to start reviewing this soon. On a related note, I'll be happy to help update this PR to the 0.6 AST if you run into any troubles. 

> Any objections to me documenting using Sphinx?

No objections. We can also add it in a follow-up if you prefer.

spookylukey:
@stasm - no problem. I've managed to update to 0.6 spec already, as per the last commit above.

I'm now getting failures on Python 2.7 only. It turns out there has been a change that `Identifier.name` is now a bytestring (`str`) object on Python 2.7 only. Before it was a unicode string, and this is causing the failure (resolver tries to use unicode strings only).

Before:

    >>> FluentParser().parse('foo = { bar }').body[0].value.elements[0].expression.id.name
    u'bar'

After 0.6 changes merged:

    >>> FluentParser().parse('foo = { bar }').body[0].value.elements[0].expression.id.name
    'bar'

 On Python 3, it was and still is a unicode string. I could work around this on Python 2.7, but this seems like an unintentional change, and doesn't match other things e.g. `MessageReference.name` etc.

stasm:
Yay, Python 2 and Unicode :)

As far as I can tell, this happens because the parser has an internal iterator over the string it parses. The iterator yields char by char. If the string passed into `FluentParser.parse` is a bytestring, then the chars yielded are bytestrings as well. For instance, this works on Python 2:

```python
>>> FluentParser().parse(u'foo = { bar }').body[0].value.elements[0].expression.id.name
u'bar'
```

I see that your tests import `unicode_literals`, so I'm not quite sure why you're running into this.

spookylukey:
@stasm - thanks for the info, I've found it - the `dedent_ftl` module wasn't using unicode literals, so was converting unicode strings to byte strings.

spookylukey:
Is there any update on this? I know it is a large amount of code to review, but it has also been quite a few months, plus I have #71 and #75 also waiting. At some point I would like to announce [django-ftl](https://github.com/django-ftl/django-ftl) but it would give a bad impression at the moment because it would depend on an unpublished version of python-fluent.

It looks like there have been changes on master which will break this branch, and I'm happy to update the code again, and I'm happy to be involved in fixing "my parts" of the code for further changes to Fluent spec, but it would be nice to know the plan.

stasm:
Sorry for not following through, @spookylukey. At no point was it my intention to leave this open and without comments for so long. I focused on Syntax 0.6 and 0.7 milestones, and now I'm working on the 0.8. You can follow the progress in our [Discourse category](https://discourse.mozilla.org/c/fluent). As an implementer, I'd love to hear your feedback on the changes planned in 0.8. It's most likely the last milestone before 1.0 which I hope to publish by the end of the year.

I'll make figuring out the review plan for this PR my priority and I'll report back by the end of the week. And BTW, `django-ftl` looks amazing!

spookylukey:
Thanks for pointing me to discourse, I should definitely follow more actively and will do so now!

stasm:
> So it looks like a bug in the parser?

Thanks for catching this. It's an artifact of the optimistic parser which landed in the runtime `fluent` package as part of the Syntax 0.7 milestone (https://github.com/projectfluent/fluent.js/pull/289).

This test should use a term attribute rather than a message attribute. For term attributes, the idea is to allow selecting based on grammatical properties of terms, e.g. `-brandname.gender`. We haven't found good use-cases for using message attributes as selectors, though. In the future, we'll provide a linter which will help translation authors avoid such mistakes. For the time being, we made it a syntax error to ensure no valid Fluent file relies on this behavior. (Plus, we can always relax it later.)

I filed https://github.com/projectfluent/fluent.js/issues/301 to a) fix the test in `fluent.js`, and b) document the optimistic parser behavior with more tests like the one you found.

stasm:
Thanks again, @spookylukey, for putting so much effort and work into this PR. It's a pleasure to read the code and documentation. I'm still reviewing it, and I'd like to propose the following landing plan.

The current `fluent` package on PyPI is used in production for building Firefox. It only provides the parser module which is used to validate translations shipping in Firefox. To ensure maximum flexibility, we think it would be smart to version it separately from the bundle implementation from this PR. (`MessageContext` has been renamed to `FluentBundle` recently.) By versioning them separately, we'll be able to make changes to the parser on a timeline required by the shipping schedule of Firefox. In practice, this flexibility will continue to be important only for a couple more months, until we hit the 1.0 milestone. After 1.0 the Fluent Syntax spec will only change in a fully backwards-compatible way (compatible with all 1.x versions) making future updates easier. For the time being, a lot of coordination is required to make sure implementations in the runtime and in all tools align and can parse the same versions of the syntax.

This repository currently uses the structure suggested by https://packaging.python.org/guides/packaging-namespace-packages/ which allows other distribution packages to use the same `fluent` package namespace. We take advantage of this to make the [`fluent-migration`](https://hg.mozilla.org/l10n/fluent-migration) module available as `flunt.migrate`. I'd like to suggest that we land the bundle implementation in a new repository in the `projectfluent` org on GitHub, use the package-namespace directory layout there too, and publish it as a separate distribution package on PyPI. We might also want to rename this repository, as well as the current `fluent` package on PyPI, to better reflect the fact that it only contains the parser implementation. I'm open to naming suggestions. For repos, do you think `python-fluent-syntax` and `python-fluent-bundle` would be suitable? Or maybe `fluent-syntax.py` and `fluent-bundle.py`? Are there strong conventions in naming repos hosting Python projects? For PyPI, `fluent-syntax` and `fluent-bundle` might be good choices, unless there's a way to group all published packages under a single org name, similar to @-scopes in npm? I'm not very familiar with good practices for publishing Python packages, and I'd love to get your advice on this.

I'd be most happy if you agreed to be the maintainer of this new repo and its corresponding package on PyPI. With separate versioning, the bundle development could proceed in its own pace, tied to a specific version of the `fluent-syntax` dependency. Most code which lands in Mozilla repos is reviewed before it's merged. Both @Pike and I will be happy to serve as first reviewers of changes landing in the new repo.

As far as this PR goes, I suggest that we keep it open for now so that we have a space to discuss the code. Expect review comments from @Pike, @zbraniecki and myself next week. When the review is completed, let's close this PR and land the code in the new repo.

Let me know if this sounds like a good plan. Thanks again!

stasm:
> I'd like to suggest that we land the bundle implementation in a new repository in the projectfluent org on GitHub, use the package-namespace directory layout there too, and publish it as a separate distribution package on PyPI. 

Or we could land it here but change the directory structure such that each distribution lives in its own directory, together with its `setup.py`:

```
fluent-syntax
    setup.py
    fluent
        __init__.py
        syntax
            __init__.py
            ...
fluent-bundle
    setup.py
    fluent
        __init__.py
        bundle
            __init__.py
            ...
```

I _think_ it would allow us to publish each distribution independently, e.g. `cd fluent-syntax; python setup.py ...`.

fluent/resolver.py
```diff
@@ -0,0 +1,391 @@
...
+@attr.s
+class ResolverEnvironment(object):
+    context = attr.ib()
+    args = attr.ib()
+    errors = attr.ib()
+    dirty = attr.ib(factory=set)
+    part_count = attr.ib(default=0)
+
+
+def resolve(context, message, args, errors=None):
```
spookylukey: After a while away, I have my own comment here!

Instead of passing `context` to resolve, I think it would probably be nicer if we passed the things from `MessageContext` that are actually needed, namely:
* the message and term dicts, possibly combined into a single dict.
* locale
* use_isolating
* functions

This would be a bit more verbose, but it would have the effect that the resolver code no longer needs to access internals of `MessageContext` (e.g. `._babel_locale`). Also `plural_form_for_number` would then come off `MessageContext` (which isn't a great home for it) and go somewhere better.

It would also make this code similar to the equivalent code in the [compiler implementation](https://github.com/projectfluent/python-fluent/pull/71/files#diff-f1879dcf9535775ee300cd035561340dR153)

fluent/context.py
```diff
@@ -0,0 +1,99 @@
...
+            msg = self._messages[name]
+            for attribute in msg.attributes:
+                if attribute.id.name == attr_name:
+                    return attribute.value
+            raise LookupError(message_id)
+        else:
+            return self._messages[message_id]
+
+    @cachedproperty
+    def plural_form_for_number(self):
```
spookylukey: This doesn't seem a great place for this function, see my comment here - https://github.com/projectfluent/python-fluent/pull/67/files#r231033619

setup.py
```diff
@@ -1,7 +1,14 @@
...
+import sys
 
 from setuptools import setup
 
+if sys.version_info < (3, 4):
+    # functools.singledispatch is in stdlib from Python 3.4 onwards.
+    extra_requires = ['singledispatch>=3.4']
+else:
+    extra_requires = []
+
```
spookylukey: I think this will give us issues in creating wheels - we'll need different wheels for Python 2, Python 3 < 3.4 and Python 3 >= 3.4 due to different requirements, which is a bit annoying, but possible, as per [PEP 425](https://www.python.org/dev/peps/pep-0425/).

spookylukey:
@stasm I'm not experienced with creating namespace packages, so not sure what to recommend. Both of the options you gave seem sensible to me.

I initially leant to having a single repo with two setup.py files. However, if we want to have independent development, then it is annoying if PRs for the parser get snagged on the fact that the build now fails due to `FluentBundle` breaking. We probably also want docs to be separate - most people who want to use `FluentBundle` do not actually care about the parser and serializer, and vice versa.  So I think we will have less friction with the rest of the tool ecosystem (Travis, Read The Docs etc) if we have separate repos, and that is *probably* the right way to go.

Regarding repo names, including `.py` in the name is not a normal convention, I think we should avoid that. `python-fluent-syntax` and `python-fluent-bundle` seem good names for repos, and `fluent-syntax` and `fluent-bundle` for packages, you often see similar things. I thought about `python-fluent-parser`, but there is also the serializer and AST definition, so `syntax` seems good, we should just make sure the package short description contains "parser" because I think it that is what most people will search for.

stasm:
> However, if we want to have independent development, then it is annoying if PRs for the parser get snagged on the fact that the build now fails due to FluentBundle breaking.

Would this still be a problem if the bundle package always used published versions of the syntax package?

> We probably also want docs to be separate - most people who want to use FluentBundle do not actually care about the parser and serializer, and vice versa.

We could still have `docs/bundle` and `docs/syntax`. They'd be hosted under the same RTD subdomain, but I see that as a feature. It will make linking and housekeeping easier.

I think I'm leaning towards keeping everything in a single repository. We already do this with `fluent.js` and I find it convenient that all issues live in the same place. This approach also makes it easier to create milestones and GitHub projects which comprise issues from all affected packages. For instance, we could set up a single short-lived project board tracking the implementation of Syntax 0.8 in both packages, or decide to use two project boards, one for each project.

It's also possible to use a [`CODEOWNERS`](https://help.github.com/articles/about-codeowners/) file in the single repo scenario to define who is maintaining which directory, and automate review requests.

Lastly, it's easier to perform housekeeping tasks in a single repo. Updating READMEs, linting code, etc. can all be done with a single PR.

spookylukey:
> Would this still be a problem if the bundle package always used published versions of the syntax package?

That's a good solution, I hadn't thought of that. The other things you mentioned sound fine, if you think it would be easier overall to have a single repo let's go with that. In fact if we decide to split at some point we can still do that, and it will be easier than trying to merge I imagine.

spookylukey:
@Pike - regarding performance, I've done almost nothing in this branch, because I put all my effort into the [compiler implementation](https://github.com/projectfluent/python-fluent/pull/71) instead - this is always going to be way faster, and if you want a fast implementation you should be using that instead. For the interpreter/evaluator clarity is most important IMO. Of course if there are easy performance gains that we should implement those - a single special case for plain text messages at the entry point probably wouldn't hurt clarity much.

README.md
```diff
@@ -30,6 +30,217 @@ you're a tool author you may be interested in the formal [EBNF grammar][].
...
+that you avoid problems down the road when non-ASCII characters start appearing
+in FTL files. If you are using Python 2, for the following examples you will
+first need to do:
+
+    >>> from __future__ import unicode_literals
+
+or add unicode literal markers to strings.
+
+To generate translations from this Python libary, you start with the
+`MessageContext` class:
```
zbraniecki: It would be cool to update the naming scheme to the current one, but I'm ok doing that in a follow up.

fluent/context.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __init__(self, locales, functions=None, use_isolating=True):
+        self.locales = locales
+        _functions = BUILTINS.copy()
+        if functions:
+            _functions.update(functions)
+        self._functions = _functions
+        self._use_isolating = use_isolating
+        self._messages = {}
+        self._terms = {}
```
zbraniecki: In other implementations we use a single hashmap for functions/messages/terms. I would be happy to see that replicated here for the sake of compliance.

fluent/context.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def add_messages(self, source):
+        parser = FluentParser()
+        resource = parser.parse(source)
+        # TODO - warn if items get overwritten
```
zbraniecki: actually, we reject items if they would override already existing element, and we report the result.

https://github.com/projectfluent/fluent-rs/blob/master/fluent/src/bundle.rs#L141

I would recommend looking at Rust signatures as they're stricter about what they can/should accept and return.

fluent/context.py
```diff
@@ -0,0 +1,99 @@
...
+                self._messages[item.id.name] = item
+            elif isinstance(item, Term):
+                self._terms[item.id.name] = item
+
+    def has_message(self, message_id):
+        try:
+            self._get_message(message_id)
+            return True
+        except LookupError:
+            return False
```
zbraniecki: I'm not an expert in python, but wouldn't it be faster if the `_get_message` returned `False` over throwing? Or are both optimizable comparibly in Python3?

fluent/context.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def message_ids(self):
+        """
+        Returns iterable of the message ids of the messages in this context
+        """
+        return six.iterkeys(self._messages)
```
zbraniecki: this is not part of `FluentBundle` API in JS/Rust. Why do you need it? And if you do, should we add it to FluentBundle API across implementations?

fluent/context.py
zbraniecki: yeah, I'd prefer not to put it here if possible. Happy to see it in a follow up.

zbraniecki: Maybe we could put it as a prop in the constructor?

fluent/context.py
```diff
@@ -0,0 +1,99 @@
...
+        Get the CLDR category for a given number.
+
+        >>> ctx = MessageContext(['en-US'])
+        >>> ctx.plural_form_for_number(1)
+        'one'
+        """
+        return babel.plural.to_python(self._babel_locale.plural_form)
+
+    @cachedproperty
+    def _babel_locale(self):
```
zbraniecki: Why not resolve it in the constructor?

fluent/exceptions.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import absolute_import, unicode_literals
```
zbraniecki: the file is called `errors` in JS and Rust impl.

fluent/exceptions.py
```diff
@@ -0,0 +1,15 @@
...
+    def __eq__(self, other):
+        return ((other.__class__ == self.__class__) and
+                other.args == self.args)
+
+
+class FluentReferenceError(FluentFormatError):
+    pass
+
+
+class FluentCyclicReferenceError(FluentFormatError):
```
zbraniecki: Is there a value in specializing the errors? I would think that generic `ParserError` and `ResolverError` would serve us quite far, and in your case I see you always using the string value of the error anyway for cyclic reference.
Not strongly opinionated, happy to keep it if you think it fits python better.

fluent/resolver.py
zbraniecki: I think that passing a `context` is quite safe here and I saw Stas using that convention around

I do agree that a single dict for all IDs would be useful, but instead of passing a locale, I'd recommend passing a locale fallback chain as provided to the `FluentBundle` constructor.

fluent/utils.py
```diff
@@ -0,0 +1,43 @@
+class cachedproperty(object):
```
zbraniecki: Is there a value of that over typical lazy override where you define a property as a getter that after initial execution overrides the property with its result?

zbraniecki:
I'm adding @stasm as a reviewer since he said he'd like to look at the resolve before we land this.

Stas also wanted to update the directory structure and I'm not sure if it's something we want to do before landing this (and would require rebasing) or after (I'd vote for after ;)).

README.md
```diff
@@ -30,6 +30,217 @@ you're a tool author you may be interested in the formal [EBNF grammar][].
...
+    >>> val, errs = context.format("show-total-points", {'points': 1234567})
+    >>> val
+    'You have 1,234,567 points.'
+
+
+You can specify your own formatting options on the arguments passed in by
+wrapping your numeric arguments with `fluent.types.fluent_number`:
+
+    >>> from fluent.types import fluent_number
+    >>> points = fluent_number(1234567, useGrouping=False)
```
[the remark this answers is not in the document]
spookylukey: `fluent_number` is actually a function - https://github.com/projectfluent/python-fluent/pull/67/files#diff-3e6998f08fb5d73e928f7295d4e2cfa5R205 - it normally returns an instance of `FluentNumber` which is indeed a class. It is just a small wrapper so that we can avoid creating instances unnecessarily. It can also return `FluentNone` instances if you pass `None` (I'm not sure now why or if that is necessary, I may have been following the JS implementation).

We could do clever things inside `FluentNumber.__new__` to avoid creating new instances unnecessarily, but it seems more clever than necessary, and with the `FluentNone` complication

stasm:
I opened #79 to discuss the planned changes to the directory structure. @spookylukey I'd love your input there and I'd also like to coordinate the order of landing. My understanding is that this PR needs to be rebased on top of master anyways. If we do it with the current directory structure, it will also require implementing changes introduced by [Syntax 0.8](https://github.com/projectfluent/fluent/releases/tag/v0.8.0), because that's what the `fluent.syntax` module supports right now on master. Or, we can first fix #79 which will allow `fluent.bundle` to pin an older version of `fluent.syntax` as a dependency. Syntax 0.8 can then be implemented in `fluent.bundle` on its own timeline.

fluent/context.py
```diff
@@ -0,0 +1,99 @@
...
+    Always use `MessageContext.format` to retrieve translation units from
+    a context.  Translations can contain references to other entities or
+    external arguments, conditional logic in form of select expressions, traits
+    which describe their grammatical features, and can use Fluent builtins.
+    See the documentation of the Fluent syntax for more information.
+    """
+
+    def __init__(self, locales, functions=None, use_isolating=True):
+        self.locales = locales
+        _functions = BUILTINS.copy()
```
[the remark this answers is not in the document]
spookylukey: I copy them because the next line mutates the dictionary to add in the custom ones i.e. I want a dictionary that combines the builtins and the custom ones, without mutating the builtins (or the passed in custom dictionary, for that matter, as general good practice).

`dict.copy()` does a shallow copy, so the values (which are functions) are not cloned in any way, plus the keys are immutable (strings) so they are not cloned either. So you just get a new independent dict object, but no copying of the actual keys or values.

fluent/context.py
spookylukey: I haven't timed it, but:

1. This is the kind of optimization I wouldn't usually try to second guess on Python, especially now we have PyPy to think about, which could behave quite differently (and variably depending on real life context). I suspect the PyPy JIT is helped when the non-exception case always returns the same type of object.
2. In general in Python it would be considered slightly un-idiomatic to return a sentinel for failure, because then you have to always remember to handle the sentinel, and if you don't you get a more confusing error somewhere later.
3. So, for example, other bits of the code rely on `_get_message` raising an exception (e.g. `format` implicitly relies on it, it doesn't attempt to catch the exception, and this is part of its contract).

fluent/context.py
spookylukey: OK, thanks, that is helpful.

In my later branch (which adds a compile-to-Python implementation), I have added better handling for this case, at least in terms of collecting the error  - 
https://github.com/projectfluent/python-fluent/pull/71/files#diff-f1879dcf9535775ee300cd035561340dR44

and it includes an API for actually seeing the errors (whether at the parsing or compiling stage). 

If I fix this to not overwrite, and leave the reporting of the error to my other branch (which already reports it), would that be OK?

fluent/utils.py
spookylukey: cachedproperty is quite neat in that it has zero overhead after it has run once (not even a function call, due to the way that Python look things up in object/class dicts), plus it is neat to implement. (Maybe that is what you meant by "overriding the property with its result", not sure, cachedproperty simply wraps this behaviour up neatly).

However, I don't think we really need it at all - the only use for this laziness was not really necessary.

fluent/resolver.py
spookylukey: @zbraniecki - the problem with passing a locale fallback chain is that `resolve` would then have to convert this to a Babel locale object every time it is called, in other words for every single call to `format` (even though it will always return the same thing), rather than doing it once up front.

README.md
```diff
@@ -30,6 +30,217 @@ you're a tool author you may be interested in the formal [EBNF grammar][].
...
+-----
+
+Fluent uses unicode everywhere internally, and doesn't accept bytestrings, so
+that you avoid problems down the road when non-ASCII characters start appearing
+in FTL files. If you are using Python 2, for the following examples you will
+first need to do:
+
+    >>> from __future__ import unicode_literals
+
+or add unicode literal markers to strings.
```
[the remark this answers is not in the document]
spookylukey: Good call on this, and all the other docs comments - I'll be updating the PR soon. Thanks for all the feedback.

fluent/context.py
spookylukey: This means you have to add checks for not allowing `format` to access to terms etc., but it turned out I only needed one extra check and various other things turned out shorter and easier after this change. Thanks!

fluent/exceptions.py
spookylukey: `exceptions.py` seems to be more common than `errors.py` but it looks like both have plenty of examples (according to GitHub search) and for consistency with other implementations I'll change it.

fluent/exceptions.py
spookylukey: My instinct is to leave these as specialized. I've never had a problem from overly specialized exceptions, but seen plenty of problems from overly broad - for anyone who is interested in certain types of exceptions, they end up having to look at the message itself, which is fragile, instead of the exception type. Plenty of good Python libraries define a plethora of exception types - e.g [SqlAlchemy exceptions](https://docs.sqlalchemy.org/en/latest/core/exceptions.html)

README.md
spookylukey: I've now updated it `FluentBundle`, but as part of the namespace package work - see #81 . I did that as a separate PR because all the renames will no doubt confuse this code review terribly!

stasm:
@spookylukey Should we close this in favor of #81?

spookylukey:
> @spookylukey Should we close this in favor of #81?

Yes, I was leaving it open in case there were any more comments on the code.

PR #70, opened 2018-07-18
Run tests against Python 3.7, too

stasm:


stasm:
Thanks for the pointer. I didn't know it was possible to have different OS setups in `.travis.yml`. I updated the PR.

.travis.yml
```diff
@@ -1,10 +1,19 @@
...
-python:
-  - "2.7"
-  - "3.5"
-  - "3.6"
-  - "nightly"
+matrix:
+  include:
+    - python: 2.7
+      dist: trusty
+      sudo: false
```
flodolo: (passing by, interested question) Do you really need to repeat the `sudo: false`, when it's declared at the beginning of the file? Logically, only the 3.7 one would need an override

stasm:
Ugh, I don't understand  the Travis and Tox combo. I'll be happy to cede this issue to you, @Pike, if you have ideas on how to fix it. I enabled `dist: xenial` to turn on the support for Python 3.7. I also followed https://docs.travis-ci.com/user/languages/python/#pypy-support and changed `pypy` and `pypy3` to `pypy3.5`. It looks like the test now fails because of an actual code issue?

stasm:
I also opened https://github.com/projectfluent/python-fluent/issues/89 to investigate whether we need Tox at all.

Pike:
> It looks like the test now fails because of an actual code issue?

Yes, sadly. Filed #90 on that.

stasm:
Thanks for looking into the `pypy` issue, @Pike, and for landing the fix directly on my branch.

stasm:
I merged `master` (with the fix from #90) into my branch. Travis is green now. @Pike, do we still need to specify `pypy3.5-6.0`?

Pike:
> do we still need to specify `pypy3.5-6.0`?

I'd say yes, I think it does two things for us: One, it's getting us the latest release that we'd end up with locally. The other is, it's pinning us to that release, so we're not getting back in the same boat we were in here, that a change of pypy under the hood breaks our tests, w/out actually being attributed to a pypy change.

We should also add back `pypy-6.0` in addition to `pypy3.5-6.0`

stasm:
OK, thanks for the help. I think this is good to land now. It ended up being a very simple PR :)

PR #75, opened 2018-09-10
Implemented escaping mechanism

spookylukey:
This adds a mechanism to handle escaped text (e.g. HTML escaping). The test includes an example that works with [markupsafe](https://github.com/pallets/markupsafe), and in [django-ftl](https://github.com/django-ftl/django-ftl) I'm using the same mechanism to work with HTML in Django templates.

This should be reviewed after #67 and #71 since it builds on them.


spookylukey:
Closing this for now, will re-visit later

PR #76, opened 2018-10-15
Support Syntax 0.7

stasm:
Port of https://github.com/projectfluent/fluent.js/pull/287


stasm:
@Pike Do you want to review this? It's 99% simple port of the JS version, and 1% of 0.4 compatibility. I added a short readme in `tests/syntax` to keep track of fixtures which differ between JS and Python.

Pike:
I'll look into this.

stasm:
Thanks. I have one more commit, a port of https://github.com/projectfluent/fluent.js/pull/287/commits/b09a623e794b7529b6090d38a2dc60740f0357cf, which I'll push once you're done with the review. It moves things around a bit; it would make the review hard.

fluent/syntax/parser.py
```diff
@@ -30,15 +30,15 @@ def __init__(self, with_spans=True):
         self.with_spans = with_spans
 
     def parse(self, source):
-        ps = FTLParserStream(source)
-        ps.skip_blank_lines()
+        ps = FTLParserStream(source.replace('\r\n', '\n'))
```
Pike: This will invalidate all spans including error positions for files with Windows line endings.

Looking at the js side of things, that doesn't handle Windows line endings at all right now?

I think we should use the algorithm we have in the reference parser to fix this issue, i.e., normalize newlines on text node creation.

fluent/syntax/parser.py
```diff
@@ -92,15 +92,15 @@ def parse_entry(self, source):
         Preceding comments are ignored unless they contain syntax errors
         themselves, in which case Junk for the invalid comment is returned.
         """
-        ps = FTLParserStream(source)
-        ps.skip_blank_lines()
+        ps = FTLParserStream(source.replace('\r\n', '\n'))
```
Pike: Also here.

Pike:
I'd like to see the disambiguation between a failed `takeChar` and EOF in the python part, too.

The key lines would be 

https://github.com/projectfluent/fluent.js/blob/61187c02ff8e7c08822ba7c828bf1aa9d7c4c42b/fluent-syntax/src/stream.js#L135-L142

which used to return `undefined` at the end and now return `null`. The python impl could return `False`, that might just work?

I ran the branch through compare-locales tests which pass, and also recreated a healthy amount of gecko-strings quarantine with the branch which resulted in the same hashes that flod generated with current master.

PR #77, opened 2018-11-13
Demo a flag-based tooling parser

Pike:
Stas, this would be a feedback? in old bugzilla days ;-)

A very rudimentary start at a syntax-detecting tooling parser.

I tested it locally against

```properties
# Foo
foo =  bar

[[joe]]
bar = hi
```

and master parses a `GroupComment` and this branch gives a `Junk` for the `[[joe]]` line.

Questions:
Is this something you'd like me to spend time on?
I chose a separation between features and syntaxes, to make the code easier to understand, but also pave the way for future versions where there's overlap between syntaxes. Does that make sense?

### DIFF
--- a/fluent/syntax/parser.py
+++ b/fluent/syntax/parser.py
@@ -5,6 +5,20 @@ from .stream import EOF, EOL, FluentParserStream
 from .errors import ParseError
 
 
+def bitfield():
+    i = 0
+    while True:
+        yield 2**i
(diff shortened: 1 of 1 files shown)


stasm:
With #107 merged, this PR can be closed. We ended up removing the support for old syntax from the parser and thus not having to implement the flag-based approach.

PR #78, opened 2018-12-06
Implement Syntax 0.8

stasm:
Port of https://github.com/projectfluent/fluent.js/pull/317.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 tests/syntax/fixtures_reference/crlf.ftl eol=crlf
+tests/syntax/fixtures_reference/cr.ftl eol=cr
 tests/syntax/fixtures_structure/crlf.ftl eol=crlf
--- a/fluent/syntax/ast.py
+++ b/fluent/syntax/ast.py
@@ -3,13 +3,13 @@ import sys
 import json
 
 
-def to_json(value):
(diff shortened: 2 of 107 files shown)


stasm:
@Pike, the changes in `fluent/syntax` are taken one-to-one from the JS implementation. There's one case where I needed to handle Python2 vs 3 explicitly: when unescaping escape sequences in `StringLiterals`, `fluent.js` uses `String.fromCodePoint`. In Python 2, we need `unichr`; in Python 3 that's simply `chr`.

I also realized that neither structure nor behavior tests have a test which would discover the need to use `unichr` vs. `chr`. So I added reference fixtures in a separate commit :) Two things to note:

- Just as in `fluent.js`, we need to skip two reference fixtures (`leading_dots.ftl` and `variant_lists.ftl`) due to how the tooling parsers reject the whole message if an attribute is broken.
- I added a processing function argument to `BaseNode.to_json()` so that I can remove the `span` property from the JSON object. I also use it to reset Junk annotations to `[]`.

stasm:
Thanks! I'll merge this on Monday.

The fixtures differences are documented in https://github.com/projectfluent/python-fluent/blob/master/tests/syntax/README.md. I'd like to find a solution and remove them soon after this lands. Thanks again!

Issue #79, opened 2018-12-18
Use namespace packages


stasm:
In preparation for #67 (or as a follow-up), I'd like to change the directory structure of this repository and start distributing Fluent code as [namespaced packages](https://packaging.python.org/guides/packaging-namespace-packages):

> Namespace packages allow you to split the sub-packages and modules within a single package across multiple, separate distribution packages

Here's the directory structure I suggested in https://github.com/projectfluent/python-fluent/pull/67#issuecomment-435599510:

```
fluent-syntax
    setup.py
    fluent
        __init__.py
        syntax
            __init__.py
            ...
fluent-bundle
    setup.py
    fluent
        __init__.py
        bundle
            __init__.py
            ...
```

spookylukey:
As I said before, I don't have any experience with namespace packages, but I'm happy to go with this proposal. If separate repos are a better option, we can always do that later.

I'd suggest doing this change first, before merging #67. Then, I can adapt #67 to depend on a specific, older version of fluent. This will be a test of our ability to manage two packages within one repo.

Where do tests fit into this? It seems like there should be two separate test directories. We could have tox/travis set to run them as different 'environments', so that failures in one package are distinct from those in the other.

spookylukey:
Based on my efforts in #80 and #81 - this seems to work. I've got my `FluentBundle` implementation that now depends on a specific, older version of `fluent` (which will soon be renamed to `fluent-syntax`), and tests run against that. Travis runs both sets of test suites for all environments, but separated out so that you can see which ones are failing more easily - https://travis-ci.org/projectfluent/python-fluent/builds/478585904

stasm:
Nice. I've checked out your branch for #81 and tested locally with `fluent.syntax`, `fluent.bundle` and with the unpublished `fluent.migrate` module from https://hg.mozilla.org/l10n/fluent-migration. All worked great (with some additional dependencies installed for `fluent.bundle`; I'll test again when I review #81).

stasm:
This was fixed by @spookylukey in #80. Since then, we published `fluent.syntax` 0.10 and `fluent.runtime` 0.1.

PR #80, opened 2019-01-11
New namespace package layout

spookylukey:
First stab at #79


spookylukey:
Notes - see:

* https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages
* https://github.com/pypa/sample-namespace-packages/tree/master/pkgutil#practical-usage

spookylukey:
Hmm, that's weird. I got exactly those failures first time I tried, turned out to be due to git 'fixing' up the whitespace in some of the files, since I have this in my global git config:

    [core]
      whitespace=fix,-indent-with-non-tab,trailing-space,cr-at-eol

So, I then overrode that locally:

    git config core.whitespace ''

and re did the 'git mv' commands. Now everything works locally, the tests pass. But not on Travis. So it seems that when I pushed, something else applied that same git whitespace config...

spookylukey:
Hm, it seems that git is still actually rewriting the file when I do 'git mv', but on disk it just moves the existing file, so that what you have in the repo is different from what you have in your working directory. And it doesn't tell you that when you do git status. Nice...

fluent-syntax/setup.py
```diff
@@ -2,7 +2,7 @@
 
 from setuptools import setup
 
-setup(name='fluent',
+setup(name='fluent_syntax',
```
stasm: I'd prefer to use `_` here (or `-`; practically they're the same), but unfortunately `fluent-bundle` is [already taken on PyPI](https://pypi.org/project/fluent-bundle/). I see three paths forward:

1. We go with `fluent_syntax` and find a different name for `fluent_bundle` in the same schema, e.g. `fluent_runtime`.
2. We go with `fluent_syntax` and re-use the existing `fluent` package name for `fluent_bundle`.
3. We use a different separator. According to [PEP 426](https://www.python.org/dev/peps/pep-0426/#name), periods are allowed, too. How about `fluent.syntax` here and `fluent.bundle` in #81?

Out of the three, I think I like the option no. 3 the most, but I'm open to other opinions.

spookylukey: Option 3 also looks good to me. My choice of `_` was due to following the guidelines here - https://github.com/pypa/sample-namespace-packages/tree/master/pkgutil#practical-usage - but it sounds like a period could work too. We can try that - I'm concerned the combination with namespace packages might confuse something - and if there is a problem we can think again.

PR #81, opened 2019-01-11
Implement fluent.bundle namespace package

spookylukey:
This the #67 branch merged onto #80 

(separate PR to avoid confusing the review that is ongoing on #67).


spookylukey:
@stasm We have a problem on this - I've checked and PyPI considers `fluent-bundle` and `fluent.bundle` to be the same package. (I attempted an upload). So we need a new package name for this, perhaps 'fluent.runtime`. We can still use the same internal module structure, with imports like `fluent.bundle` as before, because the module name on the existing PyPI `fluent_bundle` package is `fluent_bundle`, so there is no clash. It is just the package name that needs to change.

stasm:
I didn't realize PyPI would treat these names as equal. @Pike found https://www.python.org/dev/peps/pep-0503/#normalized-names which mentions (and mandates) this behavior. I think `fluent.runtime` will work fine for this.

.gitignore
```diff
@@ -2,3 +2,5 @@
 *.pyc
 .eggs/
 fluent.egg-info/
+fluent_syntax.egg-info/
+fluent_bundle.egg-info/
```
stasm: Rename this to `fluent_runtime`?

README.md
```diff
@@ -30,6 +30,251 @@ you're a tool author you may be interested in the formal [EBNF grammar][].
...
+PyPI also contains an old `fluent` package which is an older version of just
+`fluent.syntax`.
+
+Usage
+-----
+
+To generate translations using the ``fluent.runtime`` package, you start with
+the `FluentBundle` class:
+
+    >>> from fluent.bundle import FluentBundle
```
stasm: One more rename here: `from fluent.runtime`.

README.md
```diff
@@ -30,6 +30,251 @@ you're a tool author you may be interested in the formal [EBNF grammar][].
...
+    >>> bundle.add_messages("show-total-points = You have { $points } points.")
+    >>> val, errs = bundle.format("show-total-points", {'points': 1234567})
+    >>> val
+    'You have 1,234,567 points.'
+
+
+You can specify your own formatting options on the arguments passed in by
+wrapping your numeric arguments with `fluent.types.fluent_number`:
+
+    >>> from fluent.bundle.types import fluent_number
```
stasm: And here: `from fluent.runtime.types`.

README.md
```diff
@@ -30,6 +30,251 @@ you're a tool author you may be interested in the formal [EBNF grammar][].
...
+docs](https://projectfluent.org/fluent/guide/functions.html#datetime). However,
+currently the only supported options to `DATETIME` are:
+
+* `timeZone`
+* `dateStyle` and `timeStyle` which are [proposed
+  additions](https://github.com/tc39/proposal-ecma402-datetime-style) to the ECMA i18n spec.
+
+To specify options from Python code, use `fluent.bundle.types.fluent_date`:
+
+    >>> from fluent.bundle.types import fluent_date
```
stasm: And here as well: `from fluent.runtime.types`.

README.md
spookylukey: These ones I was thinking of leaving i.e. currently it is just the package (and things that reflect the package name, like the directory) that are renamed to `fluent.runtime`. Imports are still using `fluent.bundle`, both in the docs and internally.

README.md
spookylukey: Similarly here. In fact, if we wanted, we could also promote some of these user-facing things into higher level - e.g. `fluent.bundle.types` could go to `fluent.types`, which might make importing `fluent_number` and `fluent_date` a bit handier. We'd just have to add `fluent.types` to setup.py to enable this.

fluent.bundle/fluent/bundle/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+        resource = parser.parse(source)
+        # TODO - warn/error about duplicates
+        for item in resource.body:
+            if isinstance(item, (Message, Term)):
+                if item.id.name not in self._messages_and_terms:
+                    self._messages_and_terms[item.id.name] = item
+
+    def has_message(self, message_id):
+        try:
+            self._get_message(message_id)
```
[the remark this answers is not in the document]
spookylukey: OK, will do so!

fluent.bundle/setup.py
```diff
@@ -0,0 +1,31 @@
...
+      keywords=['fluent', 'localization', 'l10n'],
+      classifiers=[
+          'Development Status :: 3 - Alpha',
+          'Intended Audience :: Developers',
+          'License :: OSI Approved :: Apache Software License',
+          'Programming Language :: Python :: 2.7',
+          'Programming Language :: Python :: 3.5',
+      ],
+      packages=['fluent', 'fluent.bundle'],
+      install_requires=['fluent>=0.9,<0.10'] + extra_requires,
```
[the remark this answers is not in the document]
spookylukey: Yep, good catch thanks

fluent.bundle/fluent/bundle/types.py
```diff
@@ -0,0 +1,352 @@
...
+
+import attr
+import pytz
+from babel.dates import format_date, format_time, get_datetime_format, get_timezone
+from babel.numbers import NumberPattern, parse_pattern
+
+FORMAT_STYLE_DECIMAL = "decimal"
+FORMAT_STYLE_CURRENCY = "currency"
+FORMAT_STYLE_PERCENT = "percent"
+FORMAT_STYLES = set([FORMAT_STYLE_DECIMAL,
```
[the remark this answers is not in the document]
spookylukey: I'll go with `_OPTIONS` because I also had `DATE_STYLE_OPTIONS` and `TIME_STYLE_OPTIONS`, and `_SET` ties you to a `set` when it's possible another container object might be more appropriate.

spookylukey:
Hi @stasm - thanks so much for the review, all your encouragements and helpful comments.

For most of your comments I'll just address them directly and update the PR, only if I disagree or have something to add will I comment. So interpret silence as 'thank you I agree'!

README.md
spookylukey: I think you're right, we'd need that too. Your other comment convinced be that it is probably better to go with `fluent.runtime`, it would be surprising to not have `fluent.runtime` importable.

fluent.runtime/fluent/bundle/resolver.py
```diff
@@ -0,0 +1,391 @@
...
+
+@handle.register(FluentNone)
+def handle_fluent_none(none, env):
+    return none.format(env.context._babel_locale)
+
+
+@handle.register(type(None))
+def handle_none(none, env):
+    # We raise the same error type here as when a message is completely missing.
+    raise LookupError("Message body not defined")
```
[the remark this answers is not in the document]
spookylukey: I guess I had two things in mind with this behaviour. One is the behaviour if you have a typo or copy-paste error. It seemed more useful that you'd get a loud error if you are trying to use the body of a message when that body doesn't exist at all.

The other was the situation of a higher level library that wants to implement some kind of fallback to a default locale. If every error condition returns some value (e.g. "???") it becomes harder to know when to fallback.

I imagined that for this condition, it is very likely there is some kind of developer/tranlsator error (e.g. in v1 of your translation files you had a message without a message body, in v2 it gains a message body but some of the FTL files have not been updated). So in this case you would want fallback to a different locale. 

Hopefully that also helps convince that this is the right thing (though I'm certainly open to persuasion the other way).

fluent.runtime/fluent/bundle/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+        except LookupError:
+            return False
+
+    def format(self, message_id, args=None):
+        message = self._get_message(message_id)
+        if args is None:
+            args = {}
+        errors = []
+        resolved = resolve(self, message, args, errors=errors)
+        return resolved, errors
```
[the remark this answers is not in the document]
stasm: Sorry, I didn't express myself clearly :) I meant that in JS we pass `errors` due to perf reasons; re-using it saves GC cycles. But in Python, one `format` maps to one `resolve`. There's no potential for reuse (unless I misunderstand) and thus, little difference if `errors` are created in `format` (as they are right now) or in `resolve` (as I suggested in my earlier comment). I think they belong in `resolve`.

fluent.runtime/tests/format/test_arguments.py
```diff
@@ -0,0 +1,54 @@
...
+import unittest
+
+from fluent.bundle import FluentBundle
+
+from ..utils import dedent_ftl
+
+
+class TestNumbersInValues(unittest.TestCase):
+    def setUp(self):
+        self.ctx = FluentBundle(['en-US'], use_isolating=False)
```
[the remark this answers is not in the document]
stasm: Oh sure, of course. That makes sense.

spookylukey:
@stasm 

> I've been thinking about how the resolver here differs from the one I wrote for `fluent.js`. In the JS one, I tried to stick to the following rule: all handler functions always return instances of `FluentType` subclasses. (Due to performance reasons, and thanks to the built-in `toString` and `valueOf` methods, native JS strings count as `FluentTypes` as well.) All logic is performed on these instances. At the very end, they all end up in a `Pattern`, where they get formatted to strings via `toString`.

To explain how I got where I did, to the best of my memory - 

I started off trying to follow `fluent.js` quite closely (I didn't know it was you who wrote it). I think the first significant bump I hit was that in fluent.js you call `toString(ctx)` to finally convert, which works fine for JS because you can do that on strings themselves and they don't complain, they just ignore the extra context argument.

Translating that to Python turned out tricky (because `str()` does not accept an extra argument). Of course I could have added a wrapper class that had its own `toString(context)` method, but I think I probably missed the fact you were effectively optimizing away the `FluentString` wrapper by making use of the builtin `toString` method of strings, and so I got a bit stuck working out how to translate the code effectively.

So, I ended up going my own way and just following it through, and this is the way it turned out!

> I'm also not entirely sure if I understand the need for `FluentInt`, `FluentFloat`, etc. Naively, I would assume that `FluentNumber.value` can hold any number type. `match` could have specialized branches for dealing with different pair of number values. What is the benefit of using types which subclass `int`, `float` and `Decimal`? Is it so that it's easier to write custom Fluent functions? I would be OK requiring that custom function explicitly use `FluentType.value`.

I think I was mainly thinking of custom functions which can be written assuming they are dealing with normal float/int/Decimal objects etc.,. This does have the argument of elegance for the user implementing those functions I think.

I agree that it does seem strange that the `handle` function ends up dealing with types other than AST nodes, and maybe we should re-visit this. However, I think it would be good to think about that after we've had a look at my compiler implementation. I haven't thought it through, but I have a suspicion that using or subclassing native types has advantages there. The compiler and resolver share all of the `FluentInt` etc. code, and the exceptions, but not much else. For the sake of runtime compatibility with custom functions it is important that they do share whatever we end up with for wrapper numbers etc.

fluent.runtime/tests/format/test_primitives.py
```diff
@@ -0,0 +1,142 @@
...
+        self.ctx.add_messages(dedent_ftl("""
+            foo               = Foo
+            placeable-literal = { "Foo" } Bar
+            placeable-message = { foo } Bar
+            selector-literal = { "Foo" ->
+                [Foo] Member 1
+               *[Bar] Member 2
+             }
+            bar
+                .attr = Bar Attribute
```
stasm: Here's one more.

fluent.runtime/tests/format/test_primitives.py
```diff
@@ -0,0 +1,142 @@
...
+    def setUp(self):
+        self.ctx = FluentBundle(['en-US'], use_isolating=False)
+        self.ctx.add_messages(dedent_ftl("""
+            foo               = Foo
+            bar               = { foo }Bar
+
+            placeable-message = { bar }Baz
+
+            baz
+                .attr = { bar }BazAttribute
```
stasm: Another one here.

fluent.runtime/tests/format/test_attributes.py
```diff
@@ -0,0 +1,149 @@
...
+                .attr = Bar Attribute
+            baz = { foo } Baz
+            qux = { foo } Qux
+                .attr = Qux Attribute
+            ref-foo = { foo.missing }
+            ref-bar = { bar.missing }
+            ref-baz = { baz.missing }
+            ref-qux = { qux.missing }
+            attr-only
+                     .attr  = Attr Only Attribute
```
stasm: There are still a few messages across tests which use the old, equal-sign-less syntax for messages which only define attributes. Would you mind fixing them before this PR is merged, please?

stasm:
@spookylukey, I think this is ready for landing, unless you need a bit more time. Let me know if you'd like me to merge it, or feel free to merge yourself—I've sent you an invite to collaborate on this repo :)

PR #83, opened 2019-01-24
Sphinx docs

spookylukey:
You can see them live at https://fluent-runtime.readthedocs.io/en/latest/

@stasm  if you want to do something similar for `fluent.syntax`, you could copy this set of changes and adapt, or here are the steps I took:

    $ cd fluent.runtime
    $ pip install sphinx
    $ mkdir docs
    $ cd docs

    $ sphinx-quickstart

(answer questions, filling in obvious answers, mostly accepting defaults except:

* intersphinx - y
* viewcode - y

I then added install.rst and usage.rst, and added these to the TOC in index.rst. The initial version of usage.rst was created using [pandoc](https://pandoc.org/):

    pandoc -f markdown -t rst README.md 

I then modified, especially to add `.. code-block:: python` in lots of places, and changed
the heading levels

In conf.py:

* To enable viewcode extension I added

       sys.path.insert(0, os.path.abspath('..'))

*  I changed the theme to default. This results in a better theme on readthedocs (IMO - the 'alabaster' theme looks nice but lacks useful things like 'next'/'up' links)

* Added 'version' 

Then I converted CHANGELOG.md to CHANGELOG.rst using pandoc, so that it can be included in the docs, and added docs/history.rst. I can see that you might not want to do that - it makes publishing the changelog on github releases page more annoying, because github expects Markdown not ReST. 

Hopefully that's all, I may have forgotten something.

Also you then need an account on https://readthedocs.org/ if you don't have one already, and you have to import the project which is pretty easy. You can connect to GitHub automatically but don't have to, manual import is easy. I chose 'edit advanced project options' in that process.


stasm:
Thanks for the detailed how-to, @spookylukey. I'll look into this next week.

fluent.runtime/docs/index.rst
```diff
@@ -0,0 +1,26 @@
...
+   usage
+   history
+
+
+Indices and tables
+==================
+
+* :ref:`genindex`
+* :ref:`modindex`
+* :ref:`search`
```
stasm: What are these? They're either empty or 404 on https://fluent-runtime.readthedocs.io/en/latest/.

fluent.runtime/CHANGELOG.rst
```diff
@@ -0,0 +1,9 @@
+Changelog
+=========
+
+fluent.runtime 0.1 (January 21, 2019)
+-------------------------------------
+
+First release to PyPI of ``fluent.runtime``. This release contains a
+``FluentBundle`` implementation that can generate translations from FTL spec. It
+targets the Fluent 0.6 spec.
```
stasm: I think it actually supports Syntax 0.7 :) [`setup.py` lists `fluent>=0.9,<0.10`](https://github.com/projectfluent/python-fluent/blob/fluent.runtime%400.1/fluent.runtime/setup.py#L29), and `fluent` 0.9 [added parsing support for Syntax 0.7](https://github.com/projectfluent/python-fluent/releases/tag/0.9.0).

Also, can you link to the Spec release page at https://github.com/projectfluent/fluent/releases/tag/v0.7.0, please?

fluent.runtime/docs/index.rst
spookylukey: Hmm, these were created automatically by Sphinx quick-start.

* `search` isn't needed - there is a search box in the sidebar on read the docs. 
* `genindex` is a general index, described [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=genindex#special-names) in the sphinx docs. I'm guessing we're unlikely to use it.
* `modindex` is an index of modules/classes, which is currently empty because we have no actual API docs. I think API docs would be a good addition. But I'll remove it for now, I can add it back later.

PR #87, opened 2019-01-27
Create benchmark for fluent.runtime (fixes #85)

### LINKED ISSUE #85: Create some benchmarks for fluent.runtime
@spookylukey has some benchmarks on his compiler branch, I'd love to land some variant of that ahead of time. That way we have a baseline idea on how the current code performs.

I'm wondering if we should, perhaps at least in addition to what's on branch, have some less micro-benchmark-y mini-benchmark, with, say 20 strings, 1 Term reference, one plural and variable ref? Or something like that? Still pretty unrealistic at that, but maybe a tad better.

And, maybe adjust the readme to use py-spy instead of pyflame? py-spy runs on more OSes, and bundles the flamegraph stuff.

Pike:
I haven't figured out how to actually run this on travis.

I'd love if this would be extra jobs, maybe just on py2.7, py3.7, pypy2, pypy3?

Reminds me, we're not testing py3.7 at all yet, we need to. Which is partly #70.

### DIFF
--- a/fluent.runtime/tools/benchmarks/README.md
+++ b/fluent.runtime/tools/benchmarks/README.md
@@ -1,0 +1,15 @@
+To run the benchmarks, do:
+
+    $ pip install -r tools/benchmarks/requirements.txt
+    $ py.test ./tools/benchmarks/fluent_benchmark.py::TestBenchmark --benchmark-warmup=on
+
+To profile the benchmark suite, we recommend py-spy as a
+good tool. Install py-spy: https://github.com/benfred/py-spy
+
+Then do something like this to profile the benchmark. Depending on your
+platform, you might need to use `sudo`.
+
+    $ py-spy -f prof.svg -- py.test ./tools/benchmarks/fluent_benchmark.py::TestBenchmark --benchmark-warmup=off
+
+And look at prof.svg in a browser. Note that this diagram includes the fixture
+setup, warmup and calibration phases which you should ignore.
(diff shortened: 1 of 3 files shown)


Pike:
Luke, I guess you're the best to review this?

I intentionally didn't try to solve for the case that your benchmarks served, i.e., compare fluent in a hot loop to gettext in a hot loop.

I'd rather try to see what we intend to look at, and then see if there are equivalents in gettext.

Also, we should run these in automation, but I'm not yet sure how to best do that.

spookylukey:
This all looks good to me, and I was able to run the benchmarks following your instructions.

Regarding running in automation, it is relatively easy to add things to the Travis matrix without adding all combinations - https://docs.travis-ci.com/user/customizing-the-build/#explicitly-including-jobs

But I have no idea what should be added - we presumably want something that will track the performance of these things over time? Is that even possible to do that reliably using cloud services? I found an article on [benchmarking in the cloud](https://bheisler.github.io/post/benchmarking-in-the-cloud/) which suggests not.

I found this issue - https://github.com/travis-ci/travis-ci/issues/352 - but it didn't turn up anything immediately useful.

Pike:
Thanks for the review.

For automation, I think I want to wait on #70 and then add running the benchmarks as single jobs. Once that works, we can change the automation on PRs to run the benchmark on both master and PR, and compare the two. I think running the benchmark twice on the same machine is the only plausible way to get numbers to compare.

PR #88, opened 2019-01-28
Update fluent.runtime for fluent.syntax 0.10

### LINKED ISSUE #86: Update fluent.runtime to fluent.syntax 0.10
Filing an issue for @spookylukey's work in https://github.com/django-ftl/python-fluent/compare/master...bundle_fluent_syntax_08.

Luke, is this something that's ready to look at? Related, this probably wants a squash, but does it also want a split into multiple commits?

spookylukey:
Fixes #86

### DIFF
--- a/fluent.runtime/fluent/runtime/__init__.py
+++ b/fluent.runtime/fluent/runtime/__init__.py
@@ -9,6 +9,7 @@ from fluent.syntax.ast import Message, Term
 
 from .builtins import BUILTINS
 from .resolver import resolve
+from .utils import ATTRIBUTE_SEPARATOR, TERM_SIGIL, add_message_and_attrs_to_store, ast_to_id
 
 
 class FluentBundle(object):
@@ -41,34 +42,24 @@ class FluentBundle(object):
         # TODO - warn/error about duplicates
         for item in resource.body:
             if isinstance(item, (Message, Term)):
-                prefix = "-" if isinstance(item, Term) else ""
-                full_id = prefix + item.id.name
+                full_id = ast_to_id(item)
                 if full_id not in self._messages_and_terms:
-                    self._messages_and_terms[full_id] = item
+                    # We add attributes to the store to enable faster looker
+                    # later, and more direct code in some instances.
+                    add_message_and_attrs_to_store(self._messages_and_terms, full_id, item)
 
     def has_message(self, message_id):
-        if message_id.startswith('-'):
+        if message_id.startswith(TERM_SIGIL) or ATTRIBUTE_SEPARATOR in message_id:
             return False
         return message_id in self._messages_and_terms
 
     def format(self, message_id, args=None):
-        message = self._get_message(message_id)
+        if message_id.startswith(TERM_SIGIL):
+            raise LookupError(message_id)
+        message = self._messages_and_terms[message_id]
         if args is None:
             args = {}
         return resolve(self, message, args)
-
-    def _get_message(self, message_id):
-        if message_id.startswith('-'):
-            raise LookupError(message_id)
-        if '.' in message_id:
-            name, attr_name = message_id.split('.', 1)
-            msg = self._messages_and_terms[name]
-            for attribute in msg.attributes:
-                if attribute.id.name == attr_name:
-                    return attribute.value
-            raise LookupError(message_id)
-        else:
-            return self._messages_and_terms[message_id]
 
     def _get_babel_locale(self):
         for l in self.locales:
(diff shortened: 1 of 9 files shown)


spookylukey:
> I'd prefer the context manager and the sub-state to be explicit about "Variables in TermReferences". Make it non-generic? Took me a while to understand that that is what it's doing, and that it's doing only that.
> 
> I'm not fond of the serialize-to-rich-ID-and-parse-it-right-after for Term/Message/Attribute references.

Yep, one of the issues here was that I was backporting some code and some ways of doing things from later branches where they make more sense.

With regard to the first point you made, the 'current environment' concept does come in useful for other features, so it does seem more generic than for just term args. Perhaps it would help if there were comments on each element describing why if is needed? It seems like a waste to change it if I've got a good idea I'm going to change it back again later.

With the second point, I'll backport the rest of the changes which make the 'rich ID' more useful in several ways, which should clean up that ugly parsing of the rich ID you pointed out.

spookylukey:
@Pike - hopefully I've addressed most of the things you highlighted with my latest changes. There are more things in `utils.py` now, and it's not obvious yet why they don't live in `resolver.py` since that is the only place that uses them now, but they will be used by the compiler branch so I left them there.

stasm:
I have two extra comments about the rich-id logic. They don't block this PR.

A consequence of the rich-id approach is that by storing the attributes separately from the messages they belong to, it forces any formatting API to be a _pull_ API. What I mean by that is that the consumer of the library needs to know exactly the names of the attributes they request. I think this is fine for `format`, but at least in `fluent.js` we've seen use-cases for another API which we called `compound` (it hasn't landed yet). It takes a message id (no attribute paths allowed) and returns a `{value, attributes}` object. It's great for localizing DOM elements because there's only a single message lookup involved. The specifics of this API were the subject of a long discussion in https://github.com/projectfluent/fluent.js/issues/208#issuecomment-431294807 and the outcome of the discussion was that there are good reasons for making the returned `attributes` both an iterator and a map allowing lookups, thus supporting both the _pull_ and the _push_ models. I'm saying all this to point out that the rich-id approach doesn't allow the _push_ model , at least in its current form which breaks the O(1) mapping from messages to their attributes.

That said, I expect the rich-id logic to go away when we implement Syntax 0.9. If I understand correctly, it's currently used as way to simplify the handling of `CallExpressions` and `AttributeExpressions` where the parent is a `TermReference`. Without special handling, the `TermReference` would be first formatted to a string value by `handle_term_reference`, which would make it impossible to run the logic required by the `CallExpression` or the `AttributeExpression` on the `Term` AST node itself. (_Cf._ my comment about `fully_resolve` in https://github.com/projectfluent/python-fluent/pull/81#pullrequestreview-193534892.) Syntax 0.9 (due in February) will remove `CallExpressions` and `AttributeExpressions` and instead store the attributes and call args on the reference node itself. See https://github.com/projectfluent/fluent/pull/221 for more information about this and let me know if you have questions about this change.

Thanks again!

spookylukey:
> The first test suite should test that terms referenced within other terms always start with fresh `args`.
> 
> ```ini
> -foo = Foo is {$arg}
> -bar = {-foo(arg: $arg)}
> -baz = {-foo}
> ref-bar = {-bar(arg: 1)}
> ref-baz = {-bar(arg: 1)}
> ```
> 
> In the example above, `ref-bar` should format to `1` without erros while `ref-baz` should format to `Foo is arg` because `arg` hasn't been explicitly passed to `-foo`.

Is the definition of `-bar` here legal? With current master I get this:

    >>> from fluent.syntax import parse
    >>> parse("""-bar = { -foo(arg: $arg)}""").body
    [<fluent.syntax.ast.Junk at 0x7f5f05ef76a0>]

This also appears to be following the spec correctly - named arguments can be only number or string literals: 
* master https://github.com/projectfluent/fluent/blob/master/spec/fluent.ebnf#L76
* Fluent 0.8 https://github.com/projectfluent/fluent/blob/v0.8.0/spec/fluent.ebnf#L83

stasm:
Ah, my bad, you're right of course. That's https://github.com/projectfluent/fluent/issues/230 (which I just filed because the original issue, 188, used outdated syntax and nomenclature). We'll consider extending the syntax to allow `-foo(arg: $arg)` after Fluent 1.0.

How about this test?

```properties
-foo = {$a} {$b}
-bar = {-foo(b: 2)}
-baz = {-foo}
ref-bar = {-bar(a: 1)}
ref-baz = {-baz(a: 1)}
```

`ref-bar` → `a 2`
`ref-baz` → `a b`

spookylukey:
I think that's address everything you mentioned now.

PR #91, opened 2019-01-30
types.py - compatibility for PyPy3 5.10 and later

### LINKED ISSUE #90: FluentDateTime inheriting from datetime breaks self.replace on pypy >= 5.10
pypy changed how `.replace` creates a new instance, by instantiating type(self).

Which ends up being a `FluentDateTime`, which doesn't take the positional arguments that datetime does.

@spookylukey, is this something you could tackle? I managed to reproduce this in the existing test suite on pypy3.5 6.0, but I'm not sure I understand how you're using kwargs in the `__new__` and `_init`.

This blocks updating our python testing setup in #70 .

spookylukey:
Fixes #90


spookylukey:
@Pike I did test it with PyPy3 version 6.0, I can confirm it works at least on that version.

PR #93, opened 2019-02-04
Travis: Remove unused sudo:false setting

olleolleolle:
This PR removes a now-unused Travis directive.


stasm:
Thanks for PR, @olleolleolle! It had a conflict with #70, which I resolved using GitHub's interface.

PR #94, opened 2019-02-05
Force checkouts of .ftl files to use LF

stasm:


stasm:
@spookylukey A heads-up about this change. Are you on Windows? I'm not sure if this might affect your local checkout or branches somehow, but if there's any risk, I can postpone merging it.

PR #95, opened 2019-02-05
Refactor resolver into a tree of callable objects, or partially evaluated

Pike:
This is my all-test-passing state of the proposal I made in https://discourse.mozilla.org/t/python-fluent-runtime-plans/35290/10.

I've given this a try to see how this is doing before we do changes to the non-compiling API that don't perform, notably I'm looking at #92.

This isn't yet in a state where it can be reviewed, sadly, because I didn't get to remove unused code yet. I wouldn't be surprised if this could end up being the same amount of code in the end as the current dispatch resolver.

The good news is that in the template benchmark, I cut down mean and median to 50% of what they're on master right now.

### DIFF
--- a/fluent.runtime/fluent/runtime/__init__.py
+++ b/fluent.runtime/fluent/runtime/__init__.py
@@ -8,7 +8,8 @@ from fluent.syntax import FluentParser
 from fluent.syntax.ast import Message, Term
 
 from .builtins import BUILTINS
-from .resolver import resolve
+from .prepare import Compiler
+from .resolver import ResolverEnvironment, CurrentEnvironment
 from .utils import ATTRIBUTE_SEPARATOR, TERM_SIGIL, add_message_and_attrs_to_store, ast_to_id
 
 
@@ -33,6 +34,8 @@ class FluentBundle(object):
         self._functions = _functions
         self._use_isolating = use_isolating
         self._messages_and_terms = {}
+        self._compiled = {}
+        self._compiler = Compiler()
         self._babel_locale = self._get_babel_locale()
         self._plural_form = babel.plural.to_python(self._babel_locale.plural_form)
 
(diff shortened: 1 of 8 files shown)


Pike:
@spookylukey, I've done a round of self-review now (bad TV programme tonight), I'd love to get your feedback on this.

fluent.runtime/fluent/runtime/__init__.py
```diff
@@ -44,22 +47,38 @@ def add_messages(self, source):
...
         if message_id.startswith(TERM_SIGIL):
             raise LookupError(message_id)
-        message = self._messages_and_terms[message_id]
-        if args is None:
-            args = {}
-        return resolve(self, message, args)
+        fluent_args = {}
+        if args is not None:
+            for argname, argvalue in args.items():
+                fluent_args[argname] = native_to_fluent(argvalue)
```
spookylukey: nit: this looks like it could be written more succinctly using a dict comprehension

    fluent_args = ({} if args is None else
                   {argname: native_to_fluent(argvalue) for argname, argvalue in args.items()})

fluent.runtime/fluent/runtime/prepare.py
```diff
@@ -0,0 +1,53 @@
...
+        if isinstance(item, FTL.BaseNode):
+            return self.compile(item)
+        if isinstance(item, (tuple, list)):
+            return [self(elem) for elem in item]
+        return item
+
+    def compile(self, node):
+        nodename = type(node).__name__
+        if not hasattr(resolver, nodename):
+            return node
```
spookylukey: With this kind of dynamic lookup of attributes on the `resolver` module, I think it would help comprehension if:

1. The resolver module gained an `__all__` attribute for show what it is exporting clearly, including all the `BaseResolver` subclasses.
2. Somewhere next to that there was a brief explanation of how these classes are looked up.

Otherwise someone new to the code base sees the `resolver.Message` class, for instance, but can't find any other references to it or work out how it is actually used.

Pike: I added a doc string to the module and cleaned up the global namespace a bit more.

I didn't go for adding an `__all__` after all. I started doing it, and then looked at the impact it had on our code paths, and there weren't that I could see. In particular, the pre-evaluator imports the module, and the complete namespace is on the module object. The only thing that would differ is `from fluent.runtime.resolver import *`, which we're not doing. Thus the `__all__ `would mostly be a source of typos and overlooked changes.

fluent.runtime/fluent/runtime/__init__.py
Pike: I've done half of this, as I found the inline `if` to be too hard to read. I made a top-level `if`, and then the dict comprehension or an empty dict.

Pike:
Thanks for the review, I'll land the status quo.

I'd be happy to see some follow-up issues filed for things that we can improve.

PR #96, opened 2019-02-07
Add a read-only visitor pattern.

### LINKED ISSUE #68: Implement BaseNode.visit API
`BaseNode.traverse` can be costly because it clones the entire tree under the node it ran on. We should add a new method just for non-destructive visiting. This will be useful for the `equals` methods, as well as for the `compare-locales` checks.

Pike:
This includes some performance tests to compare the visitor and a
traversal. Locally, the visitor is twice as fast.

This is the most basic Visitor pattern. I've had versions that were
more generic, and performance gains vanished quickly. In particular
optionally collecting results on exit halfed the performance benefits.

If we want to factor that in, we should have an independent base class
for that.

This should fix #68, I think.

### DIFF
--- a/fluent.syntax/fluent/syntax/ast.py
+++ b/fluent.syntax/fluent/syntax/ast.py
@@ -3,6 +3,36 @@ import sys
 import json
 
 
+class Visitor(object):
+    '''Read-only visitor pattern.
+
+    Subclass this to gather information from an AST.
+    To generally define which nodes not to descend in to, overload
+    `generic_visit`.
+    To handle specific node types, add methods like `visit_Pattern`.
+    The boolean value of the returned value determines if the visitor
+    descends into the children of the given AST node.
+    '''
+    def visit(self, value):
+        if isinstance(value, BaseNode):
+            self.visit_node(value)
+        if isinstance(value, list):
+            for node in value:
+                self.visit(node)
+
+    def visit_node(self, node):
+        nodename = type(node).__name__
+        visit = getattr(self, 'visit_{}'.format(nodename), self.generic_visit)
+        should_descend = visit(node)
+        if not should_descend:
+            return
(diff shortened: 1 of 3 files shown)


fluent.syntax/fluent/syntax/ast.py
```diff
@@ -3,6 +3,35 @@
...
+            for node in value:
+                self.visit(node)
+
+    def visit_node(self, node):
+        nodename = type(node).__name__
+        entering = getattr(self, 'visit_{}'.format(nodename), self.generic_visit)
+        if not entering(node):
+            return
+        for propname, propvalue in vars(node).items():
+            self.visit(propvalue)
```
stasm: I'd like to suggest a slightly more verbose way of writing this method. It's a nit, so feel free to ignore. As you know, I've been trying to promote the idea of self-documenting code in my reviews :) Here, the value I think this brings is that it gives an explicit meaning to the return value of the `process_node` method.

```
def visit_node(self, node):
        nodename = type(node).__name__
        process_node = getattr(self, 'visit_{}'.format(nodename), self.generic_visit)
        should_continue = process_node(node)
        if should_continue:
            for propname, propvalue in vars(node).items():
                self.visit(propvalue)
```

fluent.syntax/tests/syntax/test_visitor.py
```diff
@@ -0,0 +1,116 @@
...
+            }
+        )
+
+
+class WordCounter(object):
+    def __init__(self):
+        self.word_count = 0
+    def __call__(self, node):
+        if isinstance(node, ast.TextElement):
+            self.word_count += len(node.value.split())
```
stasm: Is `split()` needed here?

stasm: D'uh, of course. Thanks :)

PR #99, opened 2019-02-12
 Add Transformer API for in-place manipulation of AST

Pike:
Also, deprecate `.traverse`.

### DIFF
--- a/fluent.syntax/fluent/syntax/ast.py
+++ b/fluent.syntax/fluent/syntax/ast.py
@@ -14,19 +14,18 @@ class Visitor(object):
     descends into the children of the given AST node.
     '''
     def visit(self, value):
-        if isinstance(value, BaseNode):
-            self.visit_node(value)
         if isinstance(value, list):
             for node in value:
                 self.visit(node)
-
-    def visit_node(self, node):
-        nodename = type(node).__name__
+            return
+        if not isinstance(value, BaseNode):
+            return
+        nodename = type(value).__name__
         visit = getattr(self, 'visit_{}'.format(nodename), self.generic_visit)
-        should_descend = visit(node)
+        should_descend = visit(value)
         if not should_descend:
             return
-        for propname, propvalue in vars(node).items():
+        for propname, propvalue in vars(value).items():
             self.visit(propvalue)
 
     def generic_visit(self, node):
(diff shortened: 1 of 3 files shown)


fluent.syntax/fluent/syntax/ast.py
```diff
@@ -14,25 +14,59 @@ class Visitor(object):
     descends into the children of the given AST node.
     '''
     def visit(self, value):
```
stasm: I think I liked the split between `visit` and `visit_node`, but it's perfectly fine to combine them into a single method. I guess this is done for symmetry with `Transformer.visit`? Also, nit: rename `value` to `node`?

fluent.syntax/fluent/syntax/ast.py
```diff
@@ -103,6 +139,21 @@ def visit(value):
...
         return fun(node)
 
+    def clone(self):
+        """Create a deep clone of the current node."""
+        def visit(value):
+            """Clone node and its descendants."""
+            if isinstance(value, BaseNode):
+                return value.clone()
+            if isinstance(value, list):
+                return [visit(child) for child in value]
```
stasm: I remember that at some point we add a `tuple` check here: https://github.com/projectfluent/python-fluent/commit/cf564c3c432c6a7a4717656df31efec81887b80f. I think the goal was to support `Annotations`' `args` field which is a tuple.

We should probably fix it in https://github.com/projectfluent/python-fluent/blob/f3f9053f0fd0650e7dfad01e61e5fcaf796fbc11/fluent.syntax/fluent/syntax/errors.py#L4-L8, but perhaps it's safer to added the check here first?

Pike: Mind blown.

https://github.com/python/cpython/blob/cfd31f0af20f76849bbec6185efea01c72349f2a/Objects/exceptions.c#L202

You cannot un-tuple `.args`.

fluent.syntax/fluent/syntax/ast.py
Pike: Right, and also to follow the pattern of `ast` more closely. What really bothered me was that `Transformer` subclasses `Visitor` (for reasons I find odd but didn't debate), and thus had `visit_node`, but couldn't use it, really.

PR #100, opened 2019-02-13
fluent.syntax 0.11.0

Pike:


fluent.syntax/CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Changelog
 
+## fluent.syntax 0.10.1 (February 13, 2019)
```
stasm: I know that technically we're not breaking any existing APIs, but numbers are cheap and I suggest we call this `0.11` to better reflect the introduction of two new APIs.

Pike: Done.

PR #105, opened 2019-03-04
Fixed dependencies

spookylukey:
* six is now a runtime requirement
* DRY tox.ini


spookylukey:
@stasm  and @Pike - I don't know how we want to work it with PRs - for small things like this is it better to just go ahead and commit (once the tests pass), or would you prefer that everything gets reviewed?

PR #108, opened 2019-03-08
Test case for regression involving single element placeables

spookylukey:
This looks like a regression @Pike  (I have another one coming too, sorry to do that to you on a Friday afternoon...!). Numbers and `FluentNone` are not getting turned into strings as they should.

I realized I didn't understand your new implementation as well as I need to, especially in terms of at what point things get resolved, so I was trying to add some specific tests and came across some issues.

Only 3 of these tests fail (those with `use_isolating=True`), the others are there for good measure.

I think you probably have a better idea how to fix these at this point in time, shall I leave them with you?

### DIFF
--- a/fluent.runtime/tests/format/test_placeables.py
+++ b/fluent.runtime/tests/format/test_placeables.py
@@ -109,3 +109,47 @@ class TestPlaceables(unittest.TestCase):
         val, errs = self.ctx.format('self-parent-ref-ok.attr', {})
         self.assertEqual(val, 'Attribute Parent')
         self.assertEqual(len(errs), 0)
+
+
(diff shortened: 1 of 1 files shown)


Pike:
So, the use-isolation one I was expecting in a different scenario, actually. When we make the transform work on the Transformer api, it will modify the AST in place, and that would be bundle-dependent.

I guess that adding and discarding `self` to a set on the context would fix this?

spookylukey:
> I guess that adding and discarding `self` to a set on the context would fix this?

I think that should do it.

Pike:
Thanks for the test cases. I've included them in #115, as well as the fix.

Issue #89, opened 2019-01-28
Don't use Tox in Travis


stasm:
According to https://docs.travis-ci.com/user/languages/python/#travis-ci-uses-isolated-virtualenvs, using both Travis and Tox isn't well supported. Travis already runs all tests from the matrix in separate virtualenvs. Can we drop Tox?

Pike:
I like testing in tox, because that's what works locally. Like, I don't see how you can test fluent-runtime locally w/out installing both fluent.syntax and fluent.runtime into a virtualenv.

spookylukey:
We could use tox for local testing and just make Travis do something different. We don't have a particularly complex setup, so the bit of duplication involved shouldn't be too hard to manage.

stasm:
Alright, this makes sense. Thank you to both of you for your input. Let's add Python 3.7 to Travis in #70 and then, let's try to drop Tox from `.travis.yml` in this issue. We'll leave Tox for local testing.

Pike:
I'm closing this, 'cause we're not using tox in travis.

Not that our `.travis.yml` is in good shape, but that's a different bug.

PR #110, opened 2019-03-11
fluent.runtime - broader dependency range for fluent.syntax

spookylukey:
We don't actually depend on anything that was added after 0.10, and a tight check gives problems with other tools (e.g. compare-locales is constrained to '>=0.10,<0.11').

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,29 @@ python:
   - "pypy3.5-6.0"
   - "nightly"
 env:
-  - PACKAGE=fluent.syntax
-  - PACKAGE=fluent.runtime
-install: pip install tox-travis
-script: cd $PACKAGE; tox
+  global:
+    - FLUENT_RUNTIME_DEFAULT_DEPS="fluent.syntax==0.12.0 attrs==19.1.0 babel==2.6.0 pytz==2018.9 six==1.12.0"
+    - FLUENT_SYNTAX_DEFAULT_DEPS="six"
+  matrix:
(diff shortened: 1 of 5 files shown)


spookylukey:
Specifying exact numbers for tox/travis has the advantage of being more deterministic. It can be annoying when Travis fails for a PR (or a push) because of a problem with a new version of a dependency, not because of anything that was changed.

On the other hand, such failures do alert you to what will actually happen for new users if they try to install the library.

Perhaps the best would be:

* Pin tests to specific versions
* Add any other version combinations we think might be problematic
* Possibly add a 'latest' combination that uses specifies only minimum version, and in Travis put that in `allow_failures`

Pike:
Sounds great, thanks for bearing with me thinking aloud on this PR.

spookylukey:
OK, this looks good to go now. We've got pinned version testing in tox and Travis, against all Python versions, with a few additional ones:

  - one run testing fluent.syntax==0.10
  - one run testing the latest of everything (what you get if you just do 'pip install fluent.runtime' with nothing else already installed).

I've checked that Travis is actually installing what I intended, and it is (apart from some cases where exact dependencies are not pinned, where it seems that 'six' and 'attrs' are already installed, and I'm not sure at what version).

After a bit of wrestling with tox-travis to do all this and failing, I removed it, so this also addresses #89, and overall it is better and simpler, even if it requires a bit of duplication.

Are we good to do on this now?

PR #112, opened 2019-03-13
.equals should not ignore Variant and Attribute order, fixes #111

### LINKED ISSUE #111: ast.BaseNode.equals shouldn't ignore order for Attributes and Variants
When talking through this API in fluent.js, we found that ignoring order in Attributes and Variants is actually not the right thing.

Backport this fix to `fluent.syntax` in python land.

Pike:
This goes back to the conversation in https://github.com/projectfluent/fluent.js/pull/172

### DIFF
--- a/fluent.syntax/fluent/syntax/ast.py
+++ b/fluent.syntax/fluent/syntax/ast.py
@@ -183,13 +183,6 @@ class BaseNode(object):
                 if len(field1) != len(field2):
                     return False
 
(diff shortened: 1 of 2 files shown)


spookylukey:
LGTM

Pike:
I removed the `sorting_key` here, in line with https://github.com/projectfluent/fluent.js/pull/172#issuecomment-474459896.

PR #113, opened 2019-03-14
0.9 port, fluent.js -> python-fluent

Pike:
This is just `fluent.syntax`.

CHANGELOG should be done on a follow-up commit on the to-be-created zeronine branch, I think.


Pike:
Tests are green, requesting review.

fluent.syntax/fluent/syntax/serializer.py
```diff
@@ -179,30 +156,39 @@ def serialize_placeable(placeable):
...
+    ):
+        return serialize_reference_expression(expression)
     if isinstance(expression, ast.SelectExpression):
         return serialize_select_expression(expression)
     if isinstance(expression, ast.Placeable):
         return serialize_placeable(expression)
     raise Exception('Unknown expression type: {}'.format(type(expression)))
 
 
+def serialize_reference_expression(expression, id=None):
```
Pike: I'm not tooooooo fond of this one, as it indicates a type weakness, I think. I've kept this in sync with the js impl, but there's no backing for this factorization in the the types in the AST.

fluent.syntax/fluent/syntax/ast.py
```diff
@@ -286,29 +283,64 @@ class Expression(SyntaxNode):
...
+            precision = len(self.value) - decimal_position - 1
+        return {
+            'value': value,
+            'precision': precision
+        }
+
 
 class MessageReference(Expression):
-    def __init__(self, id, **kwargs):
+    def __init__(self, id, attribute=None, **kwargs):
```
Pike: In the js impl, the names here are still `attr`, but with actual keyword arguments in python, these need to match the properties for `traverse` to continue to work.

fluent.syntax/fluent/syntax/ast.py
```diff
@@ -330,26 +363,11 @@ def __init__(self, selector, variants, **kwargs):
...
-
-
-class CallExpression(Expression):
-    def __init__(self, callee, positional=None, named=None, **kwargs):
-        super(CallExpression, self).__init__(**kwargs)
-        self.callee = callee
-        self.positional = positional or []
-        self.named = named or []
+class CallArguments(SyntaxNode):
+    def __init__(self, positional=None, named=None, **kwargs):
```
Pike: I didn't want to put `[]` into default arguments, as it's non-pythonic.

fluent.syntax/fluent/syntax/errors.py
```diff
@@ -60,7 +60,7 @@ def get_error_message(code, args):
     if code == 'E0025':
         return 'Unknown escape sequence: \\{}.'.format(args[0])
     if code == 'E0026':
-        return 'Invalid Unicode escape sequence: {}'.format(args[0])
+        return 'Invalid Unicode escape sequence: {}.'.format(args[0])
```
Pike: This was a difference to js, which triggered structure failures on ex-behavior tests.

fluent.syntax/fluent/syntax/serializer.py
```diff
@@ -179,30 +156,39 @@ def serialize_placeable(placeable):
...
+            expression,
+            id='-{}'.format(expression.id.name)
+        )
+    if isinstance(
+        expression, (
+            ast.MessageReference,
+            ast.FunctionReference,
+        )
+    ):
+        return serialize_reference_expression(expression)
```
Pike: This is a case each in the js impl. I opted for doing a joint `isinstance` here, but don't care much either way.

fluent.syntax/tests/syntax/test_literal.py
```diff
@@ -0,0 +1,145 @@
+# coding: utf-8
```
Pike: Note on this test, it's heavily inspired by the js tests, but I took a shortcut in the fixtures in a place or two.

Also, the `U{6}` tests don't work on standard python2, as it doesn't have wide unichar support, so I made that depend on sys.

fluent.syntax/fluent/syntax/serializer.py
stasm: Thanks for bringing it up. I'll fix this in `fluent.js` and I'll follow-up in the PR with a port of the fix.

fluent.syntax/fluent/syntax/ast.py
```diff
@@ -286,29 +283,64 @@ class Expression(SyntaxNode):
...
-class NumberLiteral(Expression):
+class Literal(Expression):
+    """An abstract base class for literals."""
     def __init__(self, value, **kwargs):
-        super(NumberLiteral, self).__init__(**kwargs)
+        super(Literal, self).__init__(**kwargs)
         self.value = value
 
+    def parse(self):
+        return {'value': self.value}
```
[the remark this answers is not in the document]
Pike: Not really. For one, `namedtuple` is kinda frowned upon. Things like `pickle` don't handle them well, etc. Probably not bad in this case.

What bothers me more is that you can't subclass `namedtuple`, so the return type of `StringLiteral` and `NumberLiteral` had no common baseclass.

`attrs` fills that gap, but I don't really want to buy that dependency for the little gains we get here. It's also soewhat absorbed by data classes, but they're a new feature in 3.7, so it's a while until we can easily use those.

fluent.syntax/fluent/syntax/serializer.py
stasm: I fixed this in `fluent.js` in https://github.com/projectfluent/fluent.js/pull/351 and I ported the fix over to this PR in https://github.com/projectfluent/python-fluent/pull/113/commits/07f762f2570c3e0ea9c5aa3e2bb453280b5cdc75.

PR #116, opened 2019-05-03
Updated for fluent.syntax 0.14+ / Fluent spec 1.0

spookylukey:


fluent.runtime/fluent/runtime/resolver.py
```diff
@@ -314,33 +291,15 @@ def __call__(self, env):
...
-            if args:
-                env.errors.append(FluentFormatError("Ignored positional arguments passed to term '{0}'"
-                                                    .format(reference_to_id(self.callee))))
-            with env.modified_for_term_reference(args=kwargs):
-                return term(env)
-
-        # builtin or custom function call
-        function_name = self.callee.id.name
+        args = [arg(env) for arg in self.arguments.positional]
+        kwargs = {kwarg.name.name: kwarg.value(env) for kwarg in self.arguments.named}
```
Pike: The `.value()` doesn't exist, right?

I didn't find tests for this, maybe we should adapt some from https://github.com/projectfluent/fluent.js/blob/master/fluent/test/macros_test.js or https://github.com/projectfluent/fluent.js/blob/master/fluent/test/values_ref_test.js ?

spookylukey: The tests for named/keyword args are here:

https://github.com/projectfluent/python-fluent/blob/master/fluent.runtime/tests/format/test_functions.py#L112

(I think I got my terminology mixed up regarding keyword/named, which is not helping things).

And for named arguments in terms:

https://github.com/projectfluent/python-fluent/blob/master/fluent.runtime/tests/format/test_parameterized_terms.py#L11

`.value()` does exist - if you change it to something else you get lots of failures.

Pike: Tripped over my own code. I searched for `def value(self`, and obviously that's not found, because that's a `__call__(self`. Doh, it's been too long.

PR #120, opened 2019-08-01
Use `split` instead of `splitlines` to serialize comments (fixes #119)

### LINKED ISSUE #119: Empty comments get dropped by the serializer
A simple fragment like

```properties
## This is a group

g1 = one
g2 = two

##

not = in that group
```

doesn't serialize right. The reason is the side note in https://docs.python.org/3/library/stdtypes.html#str.splitlines, which mentions that `splitlines` returns an empty list for an empty string instead of `[""]`. Pfff.

Pike:
`splitlines` returns `[]` for empty strings, which is not what we
want, use `split` instead.

This matches what `@fluent.syntax` does on the js side.

Zibi, can you do a review on this one?

### DIFF
--- a/fluent.syntax/fluent/syntax/serializer.py
+++ b/fluent.syntax/fluent/syntax/serializer.py
@@ -64,7 +64,7 @@ class FluentSerializer(object):
 def serialize_comment(comment, prefix="#"):
     prefixed = "\n".join([
         prefix if len(line) == 0 else "{} {}".format(prefix, line)
-        for line in comment.content.splitlines(False)
+        for line in comment.content.split("\n")
     ])
     # Add the trailing line break.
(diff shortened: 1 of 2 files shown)


zbraniecki:
You may also want to test the edge case where the comment is the last token in the FTL file?

Pike:
> You may also want to test the edge case where the comment is the last token in the FTL file?

I wanted, but I did regret. You can't get the test suite to pass because of how comment whitespace is handled just a tad differently.

I also started with the js impl, where that was just a plain syntax error. It was annoying.

To stay in the scope of this bug, I just added a trailing message.

Pike:
Thanks for the review. I'll work on a bug-fix release, so we can start using this.

PR #123, opened 2019-08-05
Uplift tests

Pike:
Zibi, can you take the review on this one, too?

### DIFF
--- a/fluent.runtime/fluent/runtime/__init__.py
+++ b/fluent.runtime/fluent/runtime/__init__.py
@@ -81,7 +81,12 @@ class FluentBundle(object):
         env = ResolverEnvironment(context=self,
                                   current=CurrentEnvironment(args=fluent_args),
                                   errors=errors)
-        return [resolve(env), errors]
+        try:
+            result = resolve(env)
+        except ValueError as e:
+            errors.append(e)
+            result = '{???}'
+        return [result, errors]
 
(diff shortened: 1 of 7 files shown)


fluent.runtime/fluent/runtime/__init__.py
```diff
@@ -81,7 +81,12 @@ def format(self, message_id, args=None):
         env = ResolverEnvironment(context=self,
                                   current=CurrentEnvironment(args=fluent_args),
                                   errors=errors)
-        return [resolve(env), errors]
+        try:
+            result = resolve(env)
+        except ValueError as e:
+            errors.append(e)
+            result = '{???}'
```
zbraniecki: Drive by, but I hope we'll find a better way to display the failed selector than `???` one day. Maybe we should still take the default variant if one exists?

Pike: This one isn't about selectors at all, it's that the Fluent file tries to run a DOS on the resolver. That's a bad world to be in, so the rendered string can just be anything, really.

zbraniecki: Ah, hmm, in rust we would just display the `{elo0}` instead of resolving it. I think it's a better UX than `???`. But we can unify it in the reference, so not an issue now :)

PR #127, opened 2019-08-09
Initial commit of a pygments plugin for Fluent formatting (fixes #126)

### LINKED ISSUE #126: Create a Pygments plugin for Fluent syntax highlighting
I'm trying to make the Firefox Fluent docs better on https://firefox-source-docs.mozilla.org/, and one thing that's missing is somewhat non-broken syntax highlighting.

In particular because pygments doesn't highlight at all if there's an error, so our `properties` attempts bail out badly most of the time.

CC @flod, @zbraniecki

Pike:
Anyone interested to review this?

Looking at @flodolo @zbraniecki @spookylukey.

You can test this on local files by running

```bash
python -m fluent.pygments.cli path/to/file.ft
```

It's not the same color scheme as you'd get in Sphinx, but it's close.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ you're a tool author you may be interested in the formal [EBNF grammar][].
 Installation
 ------------
 
-python-fluent consists of two packages:
+python-fluent consists of these packages:
 
 * `fluent.syntax` - includes AST classes and parser. Most end users will not
   need this directly. Documentation coming soon!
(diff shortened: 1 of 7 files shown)


Pike:
Screenshot 2019-08-09 at 17 36 19

flodolo:
My only request would be, if it's not too complex, to have variables (placeables) and variant keys with a different color than message IDs and attribute names.

I'm using this file to test
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl

And, for example, the nested selectors becomes quite hard to parse visually.

Pike:
The braces don't actually exist in the AST, they'd require some slightly different code paths. Basically, right now, anything that's not AST nodes like `[`, `*[`, `{`, `=` is `Punctuation`. To modify that, I'd need to send special signals to the outer loop that trigger a different markup for the next non-node code segment. Possible, but not just an addition to what we have right now.

Term identifiers also don't exist. Which is why the `-` of terms is also punctuation. Also possible, but also not straight forward.

I'll land and ship the status quo, we can build from there.

PR #131, opened 2019-12-10
format_to_parts

Pike:
My local commit says WIP, so publishing this as a Draft.


spookylukey:
I'm not sure if you are wanting feedback on this from me. The code looks sensible from my brief look, I'm not sure of the motivation behind the change or whether `format_to_parts` etc. are meant to be a part of a public interface or not.

PR #136, opened 2020-03-16
Bundle API and Localization class

Pike:
This was an experimental branch to demonstrate how to implement the bundle API of other implementations, and to create a base Localization class for developers to use, and for binding-implementers to customize.

Per https://discourse.mozilla.org/t/experimental-release-of-python-fluent-runtime/47569/2, we should go ahead and merge this now.

CC @stasm @spookylukey

### DIFF
--- a/fluent.runtime/fluent/runtime/__init__.py
+++ b/fluent.runtime/fluent/runtime/__init__.py
@@ -11,6 +11,16 @@ from .builtins import BUILTINS
 from .prepare import Compiler
 from .resolver import ResolverEnvironment, CurrentEnvironment
 from .utils import native_to_fluent
+from .fallback import FluentLocalization, AbstractResourceLoader, FluentResourceLoader
+
+
+__all__ = [
+    'FluentLocalization',
+    'AbstractResourceLoader',
+    'FluentResourceLoader',
+    'FluentResource',
+    'FluentBundle',
+]
 
 
 def FluentResource(source):
--- a/fluent.runtime/fluent/runtime/__init__.py
+++ b/fluent.runtime/fluent/runtime/__init__.py
@@ -13,6 +13,11 @@ from .resolver import ResolverEnvironment, CurrentEnvironment
 from .utils import ATTRIBUTE_SEPARATOR, TERM_SIGIL, ast_to_id, native_to_fluent
 
 
+def FluentResource(source):
+    parser = FluentParser()
+    return parser.parse(source)
+
+
 class FluentBundle(object):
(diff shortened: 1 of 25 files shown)


fluent.runtime/fluent/runtime/fallback.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def format_value(self, msg_id, args=None):
+        for bundle in self._bundles():
+            if not bundle.has_message(msg_id):
+                continue
+            msg = bundle.get_message(msg_id)
+            if not msg.value:
+                continue
+            val, errors = bundle.format_pattern(msg.value, args)
+            return val
```
stasm: Maybe return the errors, too?

fluent.runtime/fluent/runtime/__init__.py
```diff
@@ -10,17 +10,34 @@
...
+    'FluentBundle',
+]
+
+
+def FluentResource(source):
+    parser = FluentParser()
+    return parser.parse(source)
 
 
 class FluentBundle(object):
```
stasm: Should we move this to its own file?

fluent.runtime/fluent/runtime/fallback.py
```diff
@@ -0,0 +1,117 @@
...
+    Generic API for Fluent applications.
+
+    This handles language fallback, bundle creation and string localization.
+    It uses the given resource loader to load and parse Fluent data.
+    """
+    def __init__(
+        self, locales, resource_ids, resource_loader,
+        use_isolating=False,
+        bundle_class=None, functions=None,
+    ):
```
stasm: Is this PR the right place to comment on the design of the `FluentLocalization` API? Would you prefer to move this discussion to Discourse perhaps?

I'd like to suggest using the pattern we use in `@fluent/react` and `@fluent/dom`, i.e. rather than pass a lot of config parameters to the constructor, we pass an iterator over bundles created by the user. This allow the user of the API to be in charge of the exact bundle class used, the options and functions used to initialize it, and the loading logic.

Pike:
The state of this branch is released and published on PYPI. It's on a branch in case it's a dead end, but it doesn't seem to be. I think we should merge this branch as is, and create follow-ups if we want to revise APIs or refactor parts.

For the expected behavior of the Localization class, discourse is probably a better platform than an implementation ticket, yeah.

Pike:
Per conversation on discourse and a chat with stasm on matrix, I went ahead and merged this.

PR #138, opened 2020-05-13
Add API docs to fluent.runtime

Pike:
This adds the entry points to the fluent.runtime to the docs.

I've abstracted the documentation from a single-person project
a bit as I go.


Pike:
With #139, we might want to change the theme to alabaster here, too.

Pike:
I expect that I'll work off a unified conf to push to gh-pages, so this is cool for now.

PR #139, opened 2020-05-13
Add docs to fluent.syntax

Pike:


fluent.syntax/docs/api/parser.rst
```diff
@@ -0,0 +1,7 @@
+Parsing
```
stasm, line 1: The file is called `parser` which shows up in the URL. Should that be `parsing.rst` instead? Or, would it make sense to just use the name of the modules verbatim for the titles of the pages?

fluent.syntax/docs/conf.py
```diff
@@ -0,0 +1,68 @@
...
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+
+
+# -- Options for HTML output -------------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'alabaster'
```
stasm, line 48: Completely personal preference, so not blocking by any means: I find the `alabaster` theme hard to read. I like the minimal look, but in this case, I wish it had more visual cues about what's code and what's prose.

Looking at the list of the [builtin themes](https://www.sphinx-doc.org/en/stable/theming.html#builtin-themes) and after testnig them locally, I might suggest `nature`, `traditional`, and `classic` instead (in that order).

fluent.syntax/docs/conf.py
```diff
@@ -0,0 +1,68 @@
...
+
+# Example configuration for intersphinx: refer to the Python standard library.
+intersphinx_mapping = {'https://docs.python.org/3/': None}
+
+# -- Options for autodoc extension -------------------------------------------
+
+import os
+import sys
+sys.path.insert(0, os.path.abspath('..'))
+autodoc_member_order = 'bysource'
```
stasm, line 68: Is this all auto-generated? If not, there's a similar-looking path setup in commented out in lines 13-15.

fluent.syntax/docs/api/ast.rst
```diff
@@ -0,0 +1,7 @@
+AST & Tooling
```
stasm, line 1: Is there a way to exlcude some symbols from being documented? E.g. the `scalars_equal` function in `ast.py`.

fluent.syntax/docs/index.rst
```diff
@@ -0,0 +1,23 @@
+Fluent Syntax
+=============
+
+The :py:mod:`fluent.syntax` library is built for tooling around Fluent. It's
+designed to parse, analyze, process, and serialize Fluent files.
+
+The :py:mod:`fluent.syntax.ast` module implements ``Visitor`` and ``Transformer``
```
stasm, line 7: I filed #141 about moving these classes to their own module.

fluent.syntax/docs/api/parser.rst
Pike: I'm torn on this. I chose the title of the page to tell readers what we're doing, not so much about what that code is.

Re URL and page names, I have another idea, and that is to move `api/index.rst` to `api.rst` or `reference.rst`, and move `parser.rst` and `serializer.rst` to the top-level, too. But to keep their include structure.

Then URLs would be `fluent.syntax/parser.html`, which is almost the module name. I toyed around calling it `fluent.syntax/api/fluent.syntax.parser.html`, but that looks weird, even w/out the `api` in there.

What do you think?

stasm: > Re URL and page names, I have another idea, and that is to move api/index.rst to api.rst or reference.rst, and move parser.rst and serializer.rst to the top-level, too. But to keep their include structure.

I like it. Since they'd be top-level, would it make sense to use the gerung forms in the file names then? We'd have `fluent.syntax/reference.html`, `fluent.syntax/parsing.html`, and  `fluent.syntax/serializing.html`.

fluent.syntax/docs/reference.rst
```diff
@@ -0,0 +1,20 @@
...
+   :members:
+
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents:
+
+   parsing
+   ast
+   serializing
```
stasm, line 20: Non-blocking suggestion: I'd make `ast` either first or last in the list.

Pike: Per matrix, we'll take another look at the ordering when we move the Visitor and Transformer.

Issue #137, opened 2020-05-13
Use github actions for automation


Pike:
Let's consider to switch to gh actions for our automation of python-fluent.

Our travis setup is rather involved, and also over-tests.

Looking at the documentation of gh actions, we'll have a ton of nice things there.

Pike:
Sadly, we can't test actions in PRs. We'll need to add workflows, and then we can iterate on the workflows. (That's by design, I learned, so PRs can't attack master)

I have a WIP (#140 ) that tries to lint. Only the code, not tests, as those don't pass. @stasm, should we start from there?

Obviously, I'll need to restore the travis config for now.

The idea would be to add GH actions until we have everything we want there, and then drop travis.

stasm:
Starting with a minimal GH Action workflow sounds good to me.

Pike:
This is done now.

PR #140, opened 2020-05-13
Use gh actions for linting

Pike:


Pike:
This should give us something to start with for #137

PR #142, opened 2020-05-14
Fix paths to lint

Pike:


stasm:
Looks like there's a number of lines over the limit in both modules.

Pike:
Nasty lil' bugger, the pick-up of `setup.cfg` changed with flake8 3.8, and I have 3.7 locally.

Pike:
Now they pass.

PR #146, opened 2020-05-15
Drop travis automation, also remove all traces.

Pike:
We really need to polish the README, but let's try this for now.

Moved the integration test dependency to '.', as the upstream
pip bug is resolved now.


Pike:
Stas, can you check if the changes to README are an incremental improvement or not?

Pike:
I think the READMEs in fluent.js are a better example of where we should go, but I didn't want to go all the way there for the sake of this PR.

PR #148, opened 2020-05-18
Update READMEs

Pike:
Our READMEs on the fluent packages are lacking.

Also convert to restructuredText to align with rest of docs.


fluent.runtime/README.rst
```diff
@@ -0,0 +1,19 @@
+fluent.runtime |fluent.runtime|
+===============================
+
+Use `Fluent`_ to localize your Python application. It comes with a ``Localization``
+class to use, based on an implementation of bundle. It uses the parser from
```
stasm, line 5: The way it's used here, `bundle` is a bit enigmatic. Perhaps mention `FluentBundle`, i.e. the class?

(I realize this is copied from the current README; if you'd prefer to fix this later rather than in a drive-by, that's fine with me.)

fluent.syntax/README.rst
```diff
@@ -0,0 +1,22 @@
+``fluent.syntax`` |fluent.syntax|
+---------------------------------
+
+Process, analyze, and transform `Fluent`_ files.
```
stasm, line 4: Perhaps explicitly mention parsing and serializing?

stasm: Or: `Read, write, and transform...`?

fluent.runtime/README.rst
Pike: Fixed this in both READMEs.

PR #149, opened 2020-05-18
Build top-level, syntax, and runtime documentation from one config

Pike:
This builds and hosts, if you use the following mammoth to rsync to a repo with gh-pages:

```
rsync -rav --include='**/_static/documentation_options.js' --exclude='fluent.*/**/_static/*' --exclude='.*' _build/python-fluent/ ....
```

The docs are generated via a script, `./scripts/build-docs python-fluent`.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,0 +1,7 @@
+
+Welcome to Python Fluent's documentation!
+=========================================
+
+The :py:mod:`python-fluent` library contains a
+`syntax <fluent.syntax/stable/>`_ and a
+`runtime <fluent.runtime/stable/>`_ library.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,7 +1,28 @@
 
-Welcome to Python Fluent's documentation!
-=========================================
+python-fluent
+=============
 
-The :py:mod:`python-fluent` library contains a
-`syntax <fluent.syntax/stable/>`_ and a
-`runtime <fluent.runtime/stable/>`_ library.
+The :py:mod:`python-fluent` project contains several packages to
+bring `Project Fluent <https://projectfluent.org>`__ to Python.
+Visit the `Github project <https://github.com/projectfluent/python-fluent>`__
+for the full list. Two packages in particular are of general interest
+and documented here.
+
+fluent.syntax
+-------------
+
+``fluent-syntax`` is the package to use for tooling, analysis, and
+processing of Fluent files.
+
+fluent.runtime
+--------------
+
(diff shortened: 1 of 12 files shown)


Pike:
Having slept over this, I'm not super happy with the quirkyness of the `rsync` command. I think `build-docs` should do that, perhaps optionally.

I'd also like to add `.nojekyll` to the build.

Maybe both under the same option? `--to-publish` ?

Also, should I already add a workflow to create the docs in automation? I think this should be a separate workflow from the library ones.

Pike:
FWIW, the state of this branch is up on https://pike.github.io/gh-stage/

docs/index.rst
```diff
@@ -0,0 +1,7 @@
+
+Welcome to Python Fluent's documentation!
+=========================================
+
+The :py:mod:`python-fluent` library contains a
+`syntax <fluent.syntax/stable/>`_ and a
+`runtime <fluent.runtime/stable/>`_ library.
```
stasm, line 7: This single sentence is short and sweet, but also almost easy to miss entirely. Would mind adding a short intro paragraph, similar to the one in the top-level README, and listing each package in a bullet point?

scripts/build-docs
```diff
@@ -0,0 +1,47 @@
...
+        if build_dir.endswith('/stable'):
+            with open(build_dir.replace('/stable', '/index.html'), 'w') as index:
+                index.write('<meta http-equiv="refresh" content="0; URL=stable/">\n')
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument('repo_name')
+    args = parser.parse_args()
+    main(args.repo_name)
```
stasm, line 47: Nit: Will there be reasons to customize the repo name?

docs/index.rst
stasm: Also, is there any way to add links to each package's docs in the sidebar?

scripts/build-docs
Pike: The repo name isn't the important part, the top-level URL is, really. Which happens to be the repo name on gh-pages.

Which is, for my stage repo, I need to use a different name than we'd use in `python-fluent`.

docs/index.rst
Pike: Figured out a way, but it's hacky. The "right" way is probably to set up intersphinx, but I want to start dealing with that once we have docs up on projectfluent, so that dependencies could resolve at least in theory.

Pike:
@stasm, this is ready for another look. Also, confirmed locally that the artifact that we build is the right thing.

PR #151, opened 2020-05-20
Refactor visitor, drop traverse

Pike:
As I'm breaking API compat in either of those changes, let's go big and bold.

Also, writing docs is much more appealing once you figured out where they should go.

### DIFF
--- a/fluent.syntax/docs/usage.rst
+++ b/fluent.syntax/docs/usage.rst
@@ -1,0 +1,140 @@
+Using fluent.syntax
+===================
+
+The ``fluent.syntax`` package provides a parser, a serializer, and libraries
+for analysis and processing of Fluent files.
+
+Parsing
+-------
+
+To parse a full resource, you can use the :py:func:`fluent.syntax.parse`
+shorthand:
+
+.. code-block:: python
+
+   from fluent.syntax import parse
+   resource = parse("""
+   ### Fluent resource comment
+
+   first = creating a { $thing }
+   second = more content
+   """)
+
+To parse a single :py:class:`fluent.syntax.ast.Message` or :py:class:`fluent.syntax.ast.Term`, use
+:py:meth:`fluent.syntax.parser.FluentParser.parse_entry`:
+
+.. code-block:: python
+
+   from fluent.syntax.parser import FluentParser
+   parser = FluentParser()
+   key_message = parser.parse_entry("""
+   ### Fluent resource comment
+
+   key = value
+   """)
+
+Serialization
+-------------
+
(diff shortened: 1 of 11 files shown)


fluent.syntax/docs/usage.rst
```diff
@@ -0,0 +1,107 @@
...
+To create Fluent syntax from AST objects, use :py:func:`fluent.syntax.serialize` or
+:py:class:`fluent.syntax.serializer.FluentSerializer`.
+
+.. code-block:: python
+
+   from fluent.syntax import serialize
+   from fluent.syntax.serialzier import FluentSerializer
+   serialize(resource)
+   serializer = FluentSerializer()
+   serializer.serialize_entry(key_message)
```
stasm, line 48: `serialize_entry` takes a full entry node, e.g. `resource.body[0]`. How about also using `serializer.serialize(resource)` here instead, to show that it's an alternative to the `serialize` helper?

stasm: Or maybe show both?

```py
serializer = FluentSerializer()
serializer.serialize(resource)
serializer.serialize_entry(message)
```

fluent.syntax/docs/usage.rst
```diff
@@ -0,0 +1,107 @@
...
+   WordCounter.count(resource)
+   WordCounter.count(key_message)
+
+In-place Modification (Transformer)
+-----------------------------------
+
+Manipulation of an AST tree can be done in-place with a subclass of
+:py:class:`fluent.syntax.visitor.Transformer`. The coding pattern matches that
+of :py:class:`visitor.Visitor`, but you can modify the node in-place.
+You can even return different types.
```
stasm, line 88:
```suggestion
You can even return different types, or remove nodes altogether.
```

fluent.syntax/docs/usage.rst
```diff
@@ -0,0 +1,107 @@
...
+   class Skeleton(visitor.Transformer):
+       def visit_TextElement(self, node):
+         node = self.generic_visit(node)
+         node.value = ''
+         return node
+
+   skeleton = Skeleton()
+   skeleton.visit(key_message)
+   WordCounter.count(key_message)
+   0
```
stasm, line 101:
```suggestion
   # Returns 0.
```

fluent.syntax/docs/usage.rst
```diff
@@ -0,0 +1,107 @@
...
+
+.. warning::
+
+   Serializing an AST tree that was created like this doesn't produce
```
stasm, line 105:
```suggestion
   Serializing an AST tree that was created like this might not produce
```

PR #154, opened 2020-08-26
Rename runtests.py to runtests

stasm:
VS Code doesn't like the fact that `fluent.syntax/runtests.py` is in fact a Bash script rather than a Python script as the extension would suggest.


stasm:
Ah, no, that's Windows (or Git?) getting confused by a symlink :/

stasm:
I filed #155 with an alternative fix.

PR #155, opened 2020-08-26
Copy fluent.runtime/runtests.py to fluent.syntax

stasm:
VS Code doesn't like the fact that `fluent.syntax/runtests.py` looks like a Bash script (it's a symlink) rather than a Python script as the extension would suggest.

### DIFF
--- a/fluent.runtime/runtests.py
+++ b/fluent.runtime/runtests.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
 
-# This file is symlinked from both fluent.runtime and fluent.syntax directories
-
 import argparse
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


Pike (commented):
Wondering, could we move the file to the top dir instead of creating a bunch of copies?

stasm:
I think that would be good, but in this PR I'd like to just unblock myself from being able to use the Python interpreter in VS Code. Two copies is still OK, so I'd suggest to factor the script out the root dir once we have maybe 3 or 4.

Pike (approved):
I'm not a big fan, but I guess this is OK. 

My quick glance says that moving the file wouldn't require changes to runtests itself, just the callsites would need to change.

stasm:
> My quick glance says that moving the file wouldn't require changes to runtests itself, just the callsites would need to change.

I tried that and it did work when I invoked `../runtests.py` manually from the CLI. Unfortunately, `tox` was no longer able to find it and I didn't investigate further.

eemeli:
Closing, as modern VS Code correctly detects the symlinked file as Python.

PR #156, opened 2020-08-27
(fluent.syntax) Don't drop to newline when serializing patterns starting with special chars

stasm:
Port of https://github.com/projectfluent/fluent.js/pull/513.


Pike (approved):
LGTM, thanks for the port.

Things I checked for in differences to JS: `elements[0]´ would behave different in js and python, but is guarded by `any([])` being false.

Different number of files is due to the abscense of a runtime parser in python.

PR #157, opened 2020-10-01
Make fluent_number obey format specifiers for currencies

Sh4rK:
Hi,

I just started to use fluent for python, and discovered that number format specifiers do not work for currencies.
Here's a small code snippet reproducing the issue:
```python
from fluent.runtime import FluentBundle, FluentResource
from fluent.runtime.types import fluent_number

bundle = FluentBundle(["en"])
resource = FluentResource("amount = { $amount }")
bundle.add_resource(resource)

msg = bundle.get_message('amount')

amount = fluent_number(1, style="currency", currency="USD", minimumFractionDigits=0)
translated, errs = bundle.format_pattern(msg.value, {"amount": amount})
print(translated)  # Prints $1.00, expected $1.
```

Let me know if there's any issue with the patch!

### DIFF
--- a/fluent.runtime/fluent/runtime/types.py
+++ b/fluent.runtime/fluent/runtime/types.py
@@ -113,7 +113,7 @@ class FluentNumber(FluentType):
         elif self.options.style == FORMAT_STYLE_CURRENCY:
             base_pattern = locale.currency_formats['standard']
             pattern = self._apply_options(base_pattern)
-            return pattern.apply(self, locale, currency=self.options.currency)
+            return pattern.apply(self, locale, currency=self.options.currency, currency_digits=False)
 
     def _apply_options(self, pattern):
         # We are essentially trying to copy the
--- a/fluent.runtime/tests/test_types.py
+++ b/fluent.runtime/tests/test_types.py
@@ -12,6 +12,14 @@ from babel import Locale
 from fluent.runtime.types import FluentDateType, FluentNumber, fluent_date, fluent_number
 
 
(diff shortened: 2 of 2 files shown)


Pike:
Interesting topic. As we're talking currencies, @stasm, do you remember if we did anything about https://github.com/projectfluent/fluent.js/pull/464 and the non-js impls?

As for this aspect of number formatting, I'm wary of exposing the number of digits to localizers for currencies. That  doesn't sound like something that should be language specific.  @stasm ?

Still, might be good to have a way for the programmer to control this.

Sh4rK:
To clarify why I opened this PR and give some input:

The project we're working on is primarily for Hungarian audiences, but we're trying to localize it to a few other languages now.

I tried currency formatting as that's something we will need, and my problem is that Hungarian Forints (HUF) get formatted with 2 decimal places. That's unusual and unnecessary for most (common) use cases, as there's no fractional HUF in common use (the smallest value coin currently in circulation is 5 HUF, that's worth less than 0.02 USD).

Now, there _are_ some areas that actually use fractional HUF, for example banking and probably other forms of accounting, and the [ISO standard that seems relevant here specifies 2 decimal places](https://en.wikipedia.org/wiki/ISO_4217#Active_codes), but in everyday life (stores, restaurants, online shops, services, etc.) it never appears.

Because of the above two different use cases I think there should be a way to customize the formatting, even if this is not the right way.

Sh4rK:
Closing for lack of interest.

PR #161, opened 2021-02-24
Drop Python 2.7, 3.5 support, add 3.6-3.9 for fluent.syntax

Pike:
This is pretty mechanic, doing Syntax first, as fluent.runtime and fluent.pygments depend on released versions of fluent.syntax.

* Update metadata and testing matrix
* Drop six
* pyupgrade --py36-plus
* Version bump and release notes

pyupgrade comes from https://github.com/asottile/pyupgrade.

I did try `--py39-plus` just for the kicks of it, and that didn't show anything more.

PR #163, opened 2021-02-24
Drop Python 2.7, 3.5 support, add 3.6-3.9 for fluent.runtime, fluent.pygments

Pike:
Much like #161 , same same but different.

I thought I'd need fluent.syntax to be released for this, but that can decide on its own if it needs six/py2 compat or not. Once it releases, we just need to expand the compat and testing constraints here.


Pike:
As you can tell, I lied just a little bit, I guess we'll be fine requiring six now, and then bumping the dependency on syntax and dropping six later, aka post-syntax-release.

@zbraniecki r?

zbraniecki (approved):
lgtm

Issue #162, opened 2021-02-24
Drop support for Python 2.7 and 3.5


Pike:
... and update our testing and compat matrix beyond 3.5, too.

This should be done syntax first, and then the other two.

Pike:
As we're breaking compat, should we also move from pgkutil-style namespace packages to native ones?

This would be a good time, as we could just generally bump mutual requirements on all three packages. Also for fluent.migration, I guess.

https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages

sandalwoodbox:
Was this resolved by https://github.com/projectfluent/python-fluent/pull/161 & https://github.com/projectfluent/python-fluent/pull/163?

eemeli:
As noted above, support for Python < 3.6 has already been dropped.

PR #164, opened 2021-03-10
Convert to native namespace packages as we are py36+, part of #162

Pike:


Pike:
Seems that native and pkgstyle as as compatible as we need, https://packaging.python.org/guides/packaging-namespace-packages/#creating-a-namespace-package has more docs.

PR #198, opened 2024-03-10
Update dependencies

eemeli:
Maintentance-y updates to the tool configs and CI actions, adding top-level `dev-requirements.txt` and `pyproject.toml`.

After merging, I intend to push a couple of commits directly to `main` applying `isort` and `black` to the codebase; they're left out of here as that'll touch pretty much all files.

The individual commits are intended to be relatively discrete steps.


eemeli:
@eviljeff, would you have the bandwidth to look at this at some point?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
+name: CI
```
KevinMind, line 1: Seems like there are many python versions in use across the different jobs. If this is intentionally, adding a comment specifying an issue link or reason why a particular version is required will make someone (maybe you) reviewing this yaml file in 6 years very appreciative :)

If otherwise you don't need different versions you could pin the version using an environment variable and specifying this across the different jobs. This is just a suggestion.

```yaml
env:
  - PYTHON_VERSION: "3.10"

jobs:
  lint:
    ...
    with:
      python-version: ${{ env.PYTHON_VERSION }}
```

Even in a scenario where you need specific versions per purpose, using a variable can also help encapsulate that purpose.

```yaml
env:
   PYTHON_VERSION_ WITH_SPECIFIC_FEATURE: "3.6"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+        python-version: [3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy3.9, pypy3.10]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: python -m pip install ./fluent.syntax ./fluent.runtime
+      - run: python -m unittest discover -s fluent.syntax
+      - run: python -m unittest discover -s fluent.runtime
+  baseline:
```
KevinMind, line 34: Is there a more descriptive name for `baseline`. I see you're running a subset of unit tests, but it's somewhat unclear what is mean by baseline? If you are also running the same fluent.runtime unit tests in the "test" job, then what is the point of having "baseline"?

Is it something to do with the versions of fluent.syntax that are supported by python version 3.6?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+    strategy:
+      matrix:
+        fluent-syntax:
+          - ./fluent.syntax
+          - fluent.syntax==0.19.0
+          - fluent.syntax==0.18.1 six
+          - fluent.syntax==0.17.0 six
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
```
KevinMind, line 45: Can you use setup-python@v5 here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
+      - run: python -m pip install -r dev-requirements.txt
+      - run: python -m pip install ./fluent.syntax ./fluent.runtime
```
KevinMind, line 18: I guess one consequence of merging the linting jobs on the GHA level is that if there is an error installing fluent.syntax you will not know if there would have also been an error in fluent.runtime.

Previously being in separate workflows would have revealed this as they could run separately and in parallel.

Does this have a performance impact on the CI run time as well? Have you considered a reusable action for setting up the environment that can then be shared across many jobs to allow centralizing configuration while still parallelizing execution?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
```
KevinMind, line 14: Have you considered adding a pip cache? you can do something like [this](https://github.com/mozilla/addons-server/blob/master/.github/actions/setup-environment/action.yml#L15) and it will cache pip dependencies across builds. Still relies on network due to running in CI but it is still quite a bit faster in my experience.

KevinMind (approved):
I don't see anything in particular that would stop me from approving. I did find a few areas where you could consider performance impact and there are some other quick wins you can add here to make the workflows more composable, concise or run faster in general.

.github/workflows/ci.yml
eemeli: Much of that range is due to inheriting a spread from before. The 3.12 for `lint` is just the current version and could probably be dropped to rely on the default provided by `ubuntu-latest`; `baseline` uses 3.6 because we still claim compatibility with that, and `documentation`'s 3.9 is slightly arbitrary, but I'm loathe to spend many cycles on it given the doc build's complexity.

.github/workflows/ci.yml
eemeli: Eh, sure, but at the cost of additional complexity and interdependency. I'm trying to keep these actions simple, and tbh I'm not too worried about one error potentially masking a later error, given that it'll be revealed after fixing the first one.

.github/workflows/ci.yml
eemeli: Dunno, didn't try. I intend to re-refactor this soon after landing this PR, by incrementing the minimum to around 3.9, which will let us get rid of this `ubuntu-20.04` runner as well. The action versions will get updated at that time as well.

.github/workflows/ci.yml
KevinMind: I would say `baseline_python` or `min_python_compatibility` would read more clearly based on what you said. This is a nit but while we are here right?

.github/workflows/ci.yml
KevinMind: 👍

.github/workflows/ci.yml
KevinMind: One line of code to save some external bandwidth but up to you. Just thought I'd point it out.

.github/workflows/ci.yml
KevinMind: You explained it excellently just now. I Just wonder if pulling that information out of this comment thread into the action definition would be helpful for someone like myself reading the code for the first time without context. Just a light suggestion to do with what you think is best.

PR #200, opened 2024-03-18
Drop support for Python < 3.9

eemeli:


eemeli:
As noted in mozilla/moz-l10n#7, we'll need to maintain Python 3.8 compatibility for now.

PR #201, opened 2024-05-31
Respect the `with_spans` parser option for annotations

leamingrad:
This PR contains a small change to stop adding span information to `Annotation` elements when `with_spans` is set to `False` for the parser.

I've pulled this into its own PR because it wasn't obvious whether this behaviour was intentional or not (I could see an argument that annotations are a bit useless without span information). If this isn't needed, I'm happy to close this PR.

Note that the broken tests are unrelated to this PR, and are fixed in https://github.com/projectfluent/python-fluent/pull/203.

### DIFF
--- a/fluent.syntax/fluent/syntax/parser.py
+++ b/fluent.syntax/fluent/syntax/parser.py
@@ -133,7 +133,8 @@ class FluentParser:
             annot = ast.Annotation(
                 err.code, list(err.args) if err.args else None, err.message
             )
-            annot.add_span(error_index, error_index)
+            if self.with_spans:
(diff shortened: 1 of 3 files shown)


eemeli (approved):
This looks sensible. Apologies for taking so long to get at this.

It looks like there was a [CI run for this](https://github.com/projectfluent/python-fluent/actions/runs/9319765948/job/25689510218) that complained on `flake`, but it's not showing up in the PR?

eemeli:
Ah, the lint problem was fixed by the force push. Still confused by the CI results not showing up here.

PR #206, opened 2025-02-01
Fix test failures on windows

studyingegret:
`test_fallback.py` uses simple string compare (through dict) to simulate files. On Windows, paths are formed with `\` instead of `/`, which causes inconsistency in the string compare, producing the failures below. (Specifically, the `\`s were created by `os.path.join` in `FluentResourceLoader`.)

Without this fix, running `python -m unittest discover -s fluent.runtime` on Windows on df5ef402be3cf61f4a16e54e490e37e9cf276c05 produced the following failures: (Some folders in paths omitted)

```
....................................................................................................................................................E.F.F.D:\***\python-fluent\fluent.runtime\fluent\runtime\types.py:361: UserWarning: FluentDateType option hour12 is not yet supported
  warnings.warn(f"FluentDateType option {k} is not yet supported")
................s...............
======================================================================
ERROR: test_bundles (tests.test_fallback.TestLocalization.test_bundles)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\***\Python\Python313\Lib\unittest\mock.py", line 1424, in patched
    return func(*newargs, **newkeywargs)
  File "D:\***\python-fluent\fluent.runtime\tests\test_fallback.py", line 34, in test_bundles
    bundle_de = next(bundles_gen)
StopIteration

======================================================================
FAIL: test_all_exist (tests.test_fallback.TestResourceLoader.test_all_exist)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\***\Python\Python313\Lib\unittest\mock.py", line 1424, in patched
    return func(*newargs, **newkeywargs)
  File "D:\***\python-fluent\fluent.runtime\tests\test_fallback.py", line 64, in test_all_exist
    self.assertEqual(len(resources_list), 1)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 0 != 1

======================================================================
FAIL: test_one_exists (tests.test_fallback.TestResourceLoader.test_one_exists)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\***\Python\Python313\Lib\unittest\mock.py", line 1424, in patched
    return func(*newargs, **newkeywargs)
  File "D:\***\python-fluent\fluent.runtime\tests\test_fallback.py", line 76, in test_one_exists
    self.assertEqual(len(resources_list), 1)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 0 != 1

----------------------------------------------------------------------
Ran 186 tests in 0.325s

FAILED (failures=2, errors=1, skipped=1)
```

Note: This pull request has lint errors, but I don't know how to fix them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,9 @@ jobs:
       - run: python -m flake8
       - run: python -m mypy fluent.syntax/fluent fluent.runtime/fluent
   test:
-    runs-on: ubuntu-latest
+    # Workaround Python 3.7 no longer supported https://github.com/actions/setup-python/issues/962#issuecomment-2414454450
+    #runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         python-version: [3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy3.9, pypy3.10]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,11 +23,10 @@ jobs:
       - run: python -m flake8
       - run: python -m mypy fluent.syntax/fluent fluent.runtime/fluent
   test:
-    # Workaround Python 3.7 no longer supported https://github.com/actions/setup-python/issues/962#issuecomment-2414454450
-    #runs-on: ubuntu-latest
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ubuntu-22.04, windows-2022]
         python-version: [3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy3.9, pypy3.10]
     steps:
       - uses: actions/checkout@v4
--- a/fluent.runtime/tests/test_utils.py
+++ b/fluent.runtime/tests/test_utils.py
@@ -31,9 +31,9 @@ class TestFileSimulate(unittest.TestCase):
         """
         if expect_contents is None:
             self.assertFalse(os.path.isfile(filename),
-                             "Expected " + filename + " to not exist.")
+                             f"Expected {filename} to not exist.")
         else:
             self.assertTrue(os.path.isfile(filename),
-                            "Expected " + filename + " to exist.")
+                            f"Expected {filename} to exist.")
             self.assertEqual(codecs.open(filename, "r", "utf-8").read(),
                              expect_contents)
--- a/fluent.runtime/tests/test_utils.py
+++ b/fluent.runtime/tests/test_utils.py
@@ -22,6 +22,9 @@ class TestFileSimulate(unittest.TestCase):
             self.assertFileIs("en\\two.txt", "Two")
             self.assertFileIs("en/three.txt", None)
             self.assertFileIs("en\\three.txt", None)
+
+            with self.assertRaises(ValueError):
+                os.path.isfile("en/")
         patch_me(10, "b")
 
     def assertFileIs(self, filename, expect_contents):
--- a/fluent.runtime/tests/test_utils.py
+++ b/fluent.runtime/tests/test_utils.py
@@ -1,0 +1,39 @@
+import unittest
+from .utils import patch_files
+import os
+import codecs
+
+
+class TestFileSimulate(unittest.TestCase):
+    def test_basic(self):
+        @patch_files({
+            "the.txt": "The",
+            "en/one.txt": "One",
+            "en/two.txt": "Two"
+        })
+        def patch_me(a, b):
+            self.assertEqual(a, 10)
+            self.assertEqual(b, "b")
+            self.assertFileIs(os.path.basename(__file__), None)
+            self.assertFileIs("the.txt", "The")
+            self.assertFileIs("en/one.txt", "One")
+            self.assertFileIs("en\\one.txt", "One")
+            self.assertFileIs("en/two.txt", "Two")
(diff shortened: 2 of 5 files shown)


studyingegret:
@eemeli I see you ran ci task and it failed. This is [a known issue](https://github.com/actions/setup-python/issues/962) and it is because Python 3.7 reached end of life so it is no longer supported on `ubuntu-latest`. I used this workaround in another branch of my fork:

```yaml
test:
  runs-on: ubuntu-latest
  # Workaround Python 3.7 no longer supported https://github.com/actions/setup-python/issues/962#issuecomment-2414454450
  runs-on: ubuntu-22.04

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -23,7 +23,9 @@ jobs:
       - run: python -m flake8
       - run: python -m mypy fluent.syntax/fluent fluent.runtime/fluent
   test:
-    runs-on: ubuntu-latest
+    # Workaround Python 3.7 no longer supported https://github.com/actions/setup-python/issues/962#issuecomment-2414454450
+    #runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         python-version: [3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy3.9, pypy3.10]
```
eemeli, line 31: If there are test failures on Windows, shouldn't we also run the test suite there?

```suggestion
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-22.04, windows-2022]
        python-version: [3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy3.9, pypy3.10]
```

eemeli (changes requested):

.github/workflows/ci.yml
studyingegret: It feels okay, but if what you care about is that tests pass, maybe there should also be tests for `normalize_path` and related file simulating code. I can add them if you want them.

studyingegret: Sorry, I don't understand what you said (what code paths did you mean).

Anyway I decide to push my work to show it first (457d6980c8e26882da92f49d9c46d15fb4b3ab9e).

My point was that it was the file simulation code that caused the test errors on a different platform, so it makes sense to test the file simulation code.

It is an opinionated change to no longer reflect files on disk, because it seems unnecessary and unused* in current tests, and may make it difficult when you want to simulate some files don't exist and there are real files which names collide with them. I'd be happy to know what you think.

\* It can be added if it is needed later

studyingegret: I'll mark this change request conversation as resolved, since I already applied the suggestions, let's continue the discussion outside. Sorry about being clumsy, this is my first PR.

fluent.runtime/tests/utils.py
```diff
@@ -1,5 +1,58 @@
...
+        i = 0
+        while i < length:
+            if path.parts[i] == ".":
+                i += 1
+            elif i < length - 1 and path.parts[i+1] == "..":
+                i += 2
+            else:
+                res_parts.append(path.parts[i])
+                i += 1
+        return "/".join(res_parts)
```
eemeli, line 36: This seems unnecessarily complicated. Couldn't we just not support `.` or `..` within paths? That seems like it would be simpler.

```suggestion
    if path.is_absolute() or "." in path.parts or ".." in path.parts:
        raise ValueError(f"Unsupported path: {path}")
    return "/".join(path.parts)
```

fluent.runtime/tests/test_utils.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        expect_contents is None: Expect file does not exist
+        expect_contents is a str: Expect file contents to match
+        """
+        if expect_contents is None:
+            self.assertFalse(os.path.isfile(filename),
+                             "Expected " + filename + " to not exist.")
+        else:
+            self.assertTrue(os.path.isfile(filename),
+                            "Expected " + filename + " to exist.")
```
eemeli, line 37:
```suggestion
            self.assertFalse(os.path.isfile(filename),
                             f"Expected {filename} to not exist.")
        else:
            self.assertTrue(os.path.isfile(filename),
                            f"Expected {filename} to exist.")
```

eemeli (changes requested):
See inline for a few final simplifications.

> My point was that it was the file simulation code that caused the test errors on a different platform, so it makes sense to test the file simulation code.

Adding the tests is fine. I meant that as the existing `test_fallback.py` tests end up using the mocking, those tests can only succeed if the mocking works as expected. Therefore, they are also implicitly testing the mocking.

> Note: In 457d698, it is my opinionated change to no longer reflect files on disk ([code](https://github.com/projectfluent/python-fluent/commit/457d6980c8e26882da92f49d9c46d15fb4b3ab9e#diff-06055cdd17fe68ba3fb005741e094b69fbaed1feb019e907de7f037552642dc8R55)), because it seems unnecessary and unused* in current tests, and may make it difficult when you want to simulate some files don't exist and there are real files which names collide with them. I'd be happy to know what you think about that.

Yeah, dropping that seems like a good idea.

> Sorry about being clumsy, this is my first PR.

No worries, welcome to open source! Honestly, if you'd not mentioned that this was your first PR ever, I'd have presumed you to be much more experienced.

I'd be very happy to review further work from you here, or elsewhere.

eemeli (approved):
Thank you for your contribution!

Issue #209, opened 2025-07-24
message attributes not retrieved


MikeOfZen:
Message attributes are not accessible via `l10n.format_value("message.attribute")` - the method returns "message.attribute" instead of the expected attribute value.

sample ftl
```
 test-button = 
     .value = Click Me
     .title = This is a test button
     .aria-label = Test button for accessibility
```

`l10n.format_value("test-button.value")` should return `"Click Me"` but instead just returns `"test-button.value"`

looking at bundle `has_message` and `get_message`, the search is by `msg_id`, but the attributes are stored in `attributes` of a message without a way of retrieveing them.

loudblow:
Same! I can't use attributes, so I'm forced to use tons of hyphen-separated messages

eemeli:
Closed by #212

PR #212, opened 2025-11-26
runtime: add attribute formatting functionality in FluentLocalization

loudblow:
Resolving issue with accessing message attributes via FluentLocalization.format_value("message.attribute")

Connected issues:
https://github.com/projectfluent/python-fluent/issues/209

### DIFF
--- a/fluent.runtime/docs/usage.rst
+++ b/fluent.runtime/docs/usage.rst
@@ -280,15 +280,16 @@ form input may have a value, but also a placeholder attribute, aria-label
 attribute, and maybe a title attribute.
 
 .. code-block:: python
+
     >>> l10n = DemoLocalization("""
-    ...     login-input = Predefined value
-    ...         .placeholder = { $email }
-    ...         .aria-label = Login input value
-    ...         .title = Type your login email
-    """)
+    ... login-input = Predefined value
+    ...     .placeholder = { $email }
+    ...     .aria-label = Login input value
+    ...     .title = Type your login email
+    ... """)
     >>> value, attributes = l10n.format_message(
-            "login-input", {"email": "email@example.com"}
-        )
+    ...     "login-input", {"email": "email@example.com"}
+    ... )
     >>> value
     'Predefined value'
     >>> attributes
@@ -297,9 +298,10 @@ attribute, and maybe a title attribute.
 You can also use the formatted message without unpacking it.
 
 .. code-block:: python
+
     >>> fmt_msg = l10n.format_message(
-            "login-input", {"email": "email@example.com"}
-        )
+    ...     "login-input", {"email": "email@example.com"}
+    ... )
     >>> fmt_msg.value
     'Predefined value'
     >>> fmt_msg.attributes
--- a/fluent.runtime/docs/usage.rst
+++ b/fluent.runtime/docs/usage.rst
@@ -275,21 +275,35 @@ above.
 Attributes
 ~~~~~~~~~~
 When rendering UI elements, it's handy to have a single translation that
-contains everything you need in one variable. For example, a Web
-Component confirm window with an OK button, a Cancel button, and a
-message.
+contains everything you need in one variable. For example, a HTML
+form input may have a value, but also a placeholder attribute, aria-label
+attribute, and maybe a title attribute.
 
 .. code-block:: python
     >>> l10n = DemoLocalization("""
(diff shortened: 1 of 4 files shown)


eemeli (changes requested):
Extending `format_value()` like this would be rather misleading, as the added functionality is specifically about formatting a message _attribute_, not its _value_.

Instead, I think we should add a new method `format_message()`, which takes the same arguments as `format_value()`, but returns a tuple of the formatted value (or `None`, if there is no value), together with a dict of all of the formatted attributes, keyed by their names.

This connects with another Fluent design choice: attributes should not be used separately, but only as a part of the whole message.

Also, we do need to include tests and a documentation update when making an API change like this.

loudblow:
Sounds fair enough, I'll try to implement that soon. Anyway, I think that added functionality could be added too (with tests and docs, of course). According to the function name it should format any kind of values, even it is an attribute of a message, because attributes also have their own values.

fluent.runtime/fluent/runtime/fallback.py
```diff
@@ -48,6 +54,35 @@ def __init__(
...
+            elif not msg.value and formatted_attrs:
+                val = None
+            else:
+                val, _errors = bundle.format_pattern(msg.value, args)
+            return FormattedMessage(
+                # Never FluentNone when format_pattern called externally
+                cast(str, val),
+                formatted_attrs if formatted_attrs else {},
+            )
+        return FormattedMessage(msg_id, {})
```
eemeli: Let's just return a `(val, formatted_attrs)` tuple rather than defining a new class.

fluent.runtime/docs/usage.rst
```diff
@@ -272,6 +272,25 @@ instances to indicate an error or missing data. Otherwise they should
 return unicode strings, or instances of a ``FluentType`` subclass as
 above.
 
+Attributes
+~~~~~~~~~~
+When rendering UI elements, it's handy to have a single translation that
+contains everything you need in one variable. For example, a Web
+Component confirm window with an OK button, a Cancel button, and a
+message.
```
eemeli: It might be better to use a signle HTML element as the example, as was done in the original comment on #209.

eemeli (changes requested):
Looks promising! See inline for detail-y comments.

fluent.runtime/fluent/runtime/fallback.py
```diff
@@ -48,6 +54,35 @@ def __init__(
         self._bundle_cache: List[FluentBundle] = []
         self._bundle_it = self._iterate_bundles()
 
+    def format_message(
+        self, msg_id: str, args: Union[Dict[str, Any], None] = None
+    ) -> FormattedMessage:
+        for bundle in self._bundles():
+            if not bundle.has_message(msg_id):
+                continue
+            msg = bundle.get_message(msg_id)
```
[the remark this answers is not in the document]
loudblow: The first reason is consistency, format_value also uses a for loop. The second reason is simplicity: the code looks linear and less bloated to those seeing it for the first time.

fluent.runtime/fluent/runtime/fallback.py
loudblow: This is a regular tuple with attributes. Some users may prefer to pass a single DTO in their code instead of passing the message and attributes separately. Using fmt_msg.message and fmt_msg.attributes seems more intuitive than fmt_msg[0] and fmt_msg[1]

fluent.runtime/fluent/runtime/fallback.py
```diff
@@ -13,6 +13,7 @@
 )
 
 from fluent.syntax import FluentParser
+from typing_extensions import NamedTuple
```
eemeli: Why not use the base version from `typing`?

eemeli (changes requested):

eemeli (changes requested):
Looks like there are some CI failures? Otherwise looks good.

eemeli (changes requested):
Still failing CI.

eemeli (approved):
Thank you for your efforts here! Once I get an approving review for #215 I'll work on getting a new release out with this included. Note that the approval doesn't need to be from anyone specifically... 😇

loudblow:
> Thank you for your efforts here! Once I get an approving review for #215 I'll work on getting a new release out with this included. Note that the approval doesn't need to be from anyone specifically... 😇

Could you create a new release, please?

PR #217, opened 2026-01-02
Switch from unittest to pytest

eemeli:
Test styles are also adapted accordingly, in part with the help of [unittest2pytest](https://pypi.org/project/unittest2pytest/).


functionzz (commented):
Is there any reason why `test_lexer.py` in `fluent.pygments/tests` still uses unittest? Other than that everything looks good.

functionzz (approved):
Looks good to me :)

PR #218, opened 2026-01-02
Allow VariableReference as NamedArgument value

eemeli:
~This will need a rebase once #217 is merged. Only the last commit is really part of this PR.~ _Edit: done._

This corresponds to the changes proposed in projectfluent/fluent#379.


fluent.runtime/tests/format/test_parameterized_terms.py
```diff
@@ -65,6 +66,13 @@ def test_fallback(self, bundle):
         assert val == "the thing"
         assert errs == []
 
+    def test_variable_named_arg(self, bundle):
+        val, errs = bundle.format_pattern(
+            bundle.get_message("thing-variable-arg").value, {"art": "indefinite"}
+        )
+        assert val == "a thing"
+        assert errs == []
+
```
functionzz, line 75: Maybe also add a test that checks for no args like for `test_functions.py`?

```python

    def test_missing_variable_named_arg(self, bundle):
        val, errs = bundle.format_pattern(
            bundle.get_message("thing-variable-arg").value, {}
        )
        assert val == "the thing"
        assert errs == [FluentReferenceError('Unknown external: art')]

```

functionzz (commented):
Mostly looks good to me but I left a few things to check over.

For [fluent.runtime/fluent/runtime/resolver.py](https://github.com/projectfluent/python-fluent/blob/724a6cf0cba068df12cb43dd9d0071eaec14f18e/fluent.runtime/fluent/runtime/resolver.py#L401), there is a suspicious definition that may need to be changed, perhaps to:

```python
class NamedArgument(FTL.NamedArgument, BaseResolver):
    name: Identifier
    value: Union[NumberLiteral, StringLiteral, FTL.VariableReference]

```

functionzz (approved):
Looks good to me.

